PR #1, opened 2015-08-15
orginized imports

ronenhamias:
remove unused imports and organized imports


antonkharenko:
+1 even so do not see much value :-P

PR #2, opened 2015-08-16
move interfaces to api packages

ronenhamias:


antonkharenko:
This is bad practice we inherited from OSGI. I want to pass it away.

PR #5, opened 2015-08-19
Improve API for cluster bootstrap, remove schema cache, add logger and a little more

antonkharenko:


cluster/src/test/java/io/servicefabric/cluster/examples/ClusterBootstrapExamples.java
```diff
@@ -0,0 +1,47 @@
+package io.servicefabric.cluster.examples;
```
ronenhamias: IMO we should move examples to separated module

antonkharenko: Yes-yes, it is going to be separate PR - I just put it somewhere for now

ronenhamias:
+1

myroslavlisniak:
I prefer to merge only squashed commits

antonkharenko:
Agree, but in this PR I have done several different things, some commits may be squashed, but most of them are separate.

PR #15, opened 2015-08-23
Initial commit examples

ronenhamias:


cluster/src/test/java/io/servicefabric/cluster/examples/ClusterBootstrapExamples.java
```diff
@@ -25,6 +27,7 @@ public static void main(String[] args) throws InterruptedException {
 		// Start cool member
 		ICluster cluster4 = Cluster.newInstance("Cool member", 4003, seedMembers).join();
 
+		List<ClusterMember> m = cluster4.members();
```
antonkharenko: Maybe move this class to examples module if it needed?

ronenhamias: no need not going to include it as examples needs reference document for them and they should demo a specific use case so i am going to break it down to many examples...

myroslavlisniak:
Please rebase your branch and squash commits

PR #19, opened 2015-08-25
Cluster and transport code clean up part2

antonkharenko:
* Remove Type Resolver and use Class.forName() instead
* Fix unstable test(s) and race conditions on TransportChannel state transitions (cause of unstable test)


cluster/src/main/java/io/servicefabric/cluster/NetworkTimer.java
```diff
@@ -13,20 +13,35 @@
...
-class Timer {
+/**
+ * It is an utility wrapper class around {@code io.netty.util.HashedWheelTimer} which simplify its usage.
+ * This timer doesn't execute scheduled job on time, but provide approximate execution and on each tick
+ * checks either there are any jobs behind the timer and execute them. In most network applications,
+ * I/O timeout does not need to be accurate. The default tick duration is 100 milliseconds.
+ *
+ * @see io.netty.util.HashedWheelTimer
+ */
+class NetworkTimer {
```
myroslavlisniak: Why did you call it NetworkTimer, it is used without any network related stuff

transport/src/main/java/io/servicefabric/transport/protocol/protostuff/ProtostuffMessageDeserializer.java
```diff
@@ -28,22 +22,34 @@ public Message deserialize(ByteBuf bb) {
...
-			Schema schema = RuntimeSchema.getSchema(clazz);
+
+		// Handle binary data
+		if (message.getDataClass() == null) {
+			return message.getData();
+		}
+
+		// Handle object data
+		try {
+			Class<?> dataClass = Class.forName(message.getDataClass());
```
myroslavlisniak: I'm worried about performance of such approach. Specially in the case if class does not exists in the system which is much slower than happy case.
This code will run on each message so can be big performance issue.
We need to introduce some caching here I think.

cluster/src/main/java/io/servicefabric/cluster/NetworkTimer.java
antonkharenko: Because of this comment
  This timer doesn't execute scheduled job on time ... In most network applications,
  I/O timeout does not need to be accurate. The default tick duration is 100 milliseconds.

But I agree - not a best name - do you have better ideas?

antonkharenko:
Failed test due to issue https://github.com/servicefabric/servicefabric/issues/10 (not in the scope of this PR)

myroslavlisniak:
+1

transport/src/main/java/io/servicefabric/transport/protocol/protostuff/BinaryMessage.java
```diff
@@ -19,6 +21,9 @@
 	@Tag(3)
 	private byte[] data;
 
+	@Tag(4)
+	private String dataClass;
```
kshchepanovskyi: `type` is better name

antonkharenko: @kshchepanovskyi Actually in next PR I want to remove this BinaryMessage and use following approach https://github.com/servicefabric/servicefabric/blob/cluster-api-part3/transport/src/main/java/io/servicefabric/transport/protocol/MessageSchema.java (here will rename dataClass to dataType) and generic message https://github.com/servicefabric/servicefabric/blob/cluster-api-part3/transport/src/main/java/io/servicefabric/transport/protocol/Message.java

PR #20, opened 2015-08-27
fix examples according to new api

ronenhamias:


antonkharenko:
Fixed examples according to changes in API here: https://github.com/servicefabric/servicefabric/pull/21

PR #21, opened 2015-08-27
Cluster and transport code clean up. Part 3

antonkharenko:
* Introduce new Message and migrate all components to it
* Clean up IGossipProtocol interface and make it work with new Message
* Clean up IClusterMembership interface
* Clean up Cluster code and ICluster interface (without JavaDocs yet)


cluster/src/main/java/io/servicefabric/cluster/ClusterMembershipQualifiers.java
```diff
@@ -1,13 +1,21 @@
...
+import io.servicefabric.transport.TransportHeaders;
 import io.servicefabric.transport.TransportMessage;
+import io.servicefabric.transport.protocol.Message;
 import rx.functions.Func1;
 
-public final class ClusterMembershipQualifiers {
-	public static final String SYNC = "pt.openapi.core.cluster/membership/sync";
-	public static final String SYNC_ACK = "pt.openapi.core.cluster/membership/syncAck";
-	public static final String GOSSIP_MEMBERSHIP = "pt.openapi.core.cluster/membership/gossip";
+final class ClusterMembershipQualifiers {
```
artem-v: Let's have single ```ClusterQualifiers``` or ```ClusterConstants``` class per entire module instead of having several little ones.

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
```diff
@@ -1,100 +1,124 @@
...
-		this.qualifier = qualifier;
+	public Message(Object data, Map<String, String> headers) {
+		setData(data);
+		setHeaders(headers);
+	}
+
+	/**
+	 * Instantiates new message with given qualifier and headers. Headers passed a sequence of key-value pairs.
+	 */
+	public Message(Object data, String... headers) {
```
artem-v: Let's have a nice builder api for messages.

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
```diff
@@ -1,100 +1,124 @@
...
 	}
 
 	/**
 	 * Instantiates new messages with given qualifier, data and correlationId.
 	 */
-	public Message(String qualifier, Object data, String correlationId) {
-		this.qualifier = qualifier;
+	public Message(Object data, Map<String, String> headers) {
+		setData(data);
+		setHeaders(headers);
```
myroslavlisniak: Why just don't inline code from this settters into constructor and use it as base constructor calling this() from others?
This also allows to make variables final

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
antonkharenko: ok, but not in the scope of this PR

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
```diff
@@ -1,100 +1,124 @@
...
+	}
+
+	void setData(Object data) {
 		this.data = data;
-		this.correlationId = correlationId;
+	}
+
+	void setHeaders(Map<String, String> headers) {
+		if (headers == null) {
+			this.headers = Collections.emptyMap();
```
artem-v: Let ```this.headers = Collections.emptyMap()``` be from very start, in initialization block. Otherwise if no one would call ```setHeaders``` this field will be null.

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
antonkharenko: Setters and empty constructor needed for protostuff schema (see MessageSchema class)

transport/src/main/java/io/servicefabric/transport/TransportHeaders.java
```diff
@@ -0,0 +1,15 @@
+package io.servicefabric.transport;
+
+/**
+ * @author Anton Kharenko
+ */
+public final class TransportHeaders {
```
artem-v: Lets have ```TransportConstants``` where we pu constants and qualifiers.

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
myroslavlisniak: Then you should remove annotation immutable from this class, because it's not true

transport/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializer.java
```diff
@@ -0,0 +1,34 @@
...
+import io.protostuff.Schema;
+import io.protostuff.runtime.RuntimeSchema;
+import io.servicefabric.transport.utils.RecyclableLinkedBuffer;
+
+public final class ProtostuffMessageSerializer implements MessageSerializer {
+
+	private static final RecyclableLinkedBuffer recyclableLinkedBuffer = new RecyclableLinkedBuffer();
+
+	// TODO [Anton Kharenko]: Move to more appropriate place
+	static {
```
artem-v: Can't we just do ```new MessageSchema()```  or ```MessageSchema.instnce``` where approriate? Instead of  depending on ```RuntimeSchema```.

antonkharenko: This class called ProtostuffMessageSerializer - so it is ok to have dependency on protostuff here.

examples/src/main/java/io/servicefabric/examples/gossip/ClusterNodeB.java
```diff
@@ -16,14 +16,11 @@
...
 	public static void main(String[] args) {
-		// Register data types (used for serialization)
-		TransportTypeRegistry.getInstance().registerType("hello/world", Greetings.class);
-		
 		// start cluster node that listen on port 3001 and point to node A as seed node
 		ICluster clusterB =  Cluster.newInstance(3001,"localhost:3000").join();
 		
 		// spread gossip
-		clusterB.gossip().spread("hello/world", new Greetings("Greetings from ClusterMember B"));
+		clusterB.gossip().spread(new Message(new Greetings("Greetings from ClusterMember B")));
```
ronenhamias: why not hide new Message inside spread method?

antonkharenko: you can have also headers... we can provide additional helper method which will accept data and automatically create message without headers, but this is sugar - I didn't focus here on sugar just core part (this PR already too big )

antonkharenko:
@ronenhamias We need to say from where this message arrived - TransportMessage basically a wrapper that contain Message and sender member from where this message arrived. 
This class should be improved, I didn't modify it in this PR, but still conceptually it will be the same - message + origin.

transport/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializer.java
artem-v: Correct, but why to store/get MessageSchema each time  if we could know its instance upfront?

artem-v:
@ronenhamias @antonkharenko  ```Message``` now provides headers map, so ```origin``` could be put there. In this case we wouldn't need ```TransportMessage```.

transport/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializer.java
antonkharenko: It is also used for deserializing Gossip (as part of GossipRequest). And RuntimeSchema already caches schema. So it is kind of convenient.

antonkharenko:
@artem-v it is string-string map - what we should put as origin - originEndpointId?

antonkharenko:
I am just thinking that it can be kind of abuse of Message if we will extend it with additional headers which were not sent by sender.

artem-v:
@antonkharenko originEndpoint as well, as it could also be wrote down as string.

artem-v:
@antonkharenko More to say on this ``origin``. This should be a collection.  Imagine we send message from A to C via some transit nodeB: ```nodeA  --> nodeB --> nodeC```.

transport/src/main/java/io/servicefabric/transport/protocol/Message.java
antonkharenko: @myroslavlisniak  removed

antonkharenko:
@artem-v I think it is very specific and I don't think it should be built into Transport. Actually how it is done if FD make sense - if you need it it can be part of the data. Currently Transport support only point to point connection and actually it should support only point to point connection so origin only one. There may be other component on top of transport which is doing this transit sends, but it will be responsibility of this component to handle paths

artem-v:
+1

myroslavlisniak:
+1 but do not forget to squash the commits

antonkharenko:
Rebased and squash

PR #22, opened 2015-08-28
Remove "local channel" support

artem-v:


antonkharenko:
+1

@myroslavlisniak Will KPI metrics tests work (possibly with rewriting) without local transport? (off-topic: maybe lets move them here)

PR #23, opened 2015-08-28
add ICluster.isLocalMember(ClusterMember member) to validate if a given

ronenhamias:
1. add ICluster.isLocalMember(ClusterMember member) to validate if a given member is local member
2. improve the examples

### DIFF
--- a/cluster/src/main/java/io/servicefabric/cluster/Cluster.java
+++ b/cluster/src/main/java/io/servicefabric/cluster/Cluster.java
@@ -218,14 +218,4 @@ public final class Cluster implements ICluster {
 		checkState(stateUpdated, "Illegal state transition from %s to %s cluster state. Expected state %s.", state.get(), update, expected);
 	}
 
-	/**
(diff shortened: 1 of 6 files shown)


cluster/src/main/java/io/servicefabric/cluster/Cluster.java
```diff
@@ -217,4 +217,15 @@ private void updateClusterState(State expected, State update) {
...
+
+	/**
+	 * check if a given member is a local member return true in case the ClusterMember is local to the cluster instance
+	 * @param ClusterMember
+	 * @return true if this cluster memeber is a local cluster memeber
+	 * @author ronen hamias
+	 */
+	@Override
+	public boolean isLocalMember(ClusterMember member) {
+		return membership().localMember().endpoint().equals(member.endpoint());
```
antonkharenko: Lets move this method on IClusterMembership interface. So from cluster it will be:

cluster.membership().isLocalMember(member)

antonkharenko:
+1

PR #25, opened 2015-08-29
Added basic benchmarks

artem-v:


benchmark/src/main/java/test/CacheReadBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+import io.servicefabric.transport.utils.memoization.ConcurrentMapMemoizer;
+
+@Fork(2)
+@State(Scope.Thread)
+@Threads(4)
+@Warmup(iterations = 5)
+@Measurement(iterations = 5)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+public class CacheReadBenchmark {
```
artem-v: Benchmark                                  Mode  Samples     Score    Error  Units
t.CacheReadBenchmark.readFromGuavaCache    avgt       10  1231.392 � 52.884  ns/op
t.CacheReadBenchmark.readFromMemoizer      avgt       10   168.989 �  5.138  ns/op

benchmark/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializerBenchmark.java
```diff
@@ -0,0 +1,109 @@
...
+import io.servicefabric.transport.TransportHeaders;
+
+@Fork(2)
+@State(Scope.Thread)
+@Threads(4)
+@Warmup(iterations = 5)
+@Measurement(iterations = 5)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+public class ProtostuffMessageSerializerBenchmark {
```
artem-v: Benchmark                                                      Mode  Samples      Score       Error  Units
i.s.t.p.ProtostuffMessageSerializerBenchmark.deser             avgt       10   1975.701 �   165.464  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.deser1k           avgt       10   2376.907 �   201.003  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.deserGossipReq    avgt       10  66163.604 � 13710.176  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.ser               avgt       10    787.852 �    26.540  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.ser1k             avgt       10    826.898 �    10.733  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.serGossipReq      avgt       10    783.361 �    13.381  ns/op

artem-v:
```
Benchmark                                                      Mode  Samples      Score      Error  Units
i.s.t.p.ProtostuffMessageSerializerBenchmark.deser             avgt       10   1889.051 �  103.509  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.deser1k           avgt       10   2244.923 �   58.437  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.deserGossipReq    avgt       10  61081.970 � 3030.942  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.ser               avgt       10    800.075 �   12.514  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.ser1k             avgt       10    815.900 �   20.773  ns/op
i.s.t.p.ProtostuffMessageSerializerBenchmark.serGossipReq      avgt       10    825.173 �  106.665  ns/op

```

benchmark/src/main/java/test/CacheReadBenchmark.java
artem-v: Here can see ```memoizer``` x10 faster than ```guava_cache```.

benchmark/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializerBenchmark.java
artem-v: Benchmark serialization/deserialization results; ser/deser is scalable despite message size, except case ```gossipReq```, strange it takes significant time comparing to others. In the test ```gossipReq``` was ```GossipRequest``` populate with x5 regular(256b) ```Gossip``` messages, so in total it was < 1k.

antonkharenko:
Fix travis

benchmark/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializerBenchmark.java
antonkharenko: 64b x 100 ~ 6k

antonkharenko: Also I think it is not representative compare since you put into message byte array data which has special processing at MessageSchema:

```
// Write data
		Object originalData = message.data();
		if (originalData != null) {
			if (originalData instanceof byte[]) {
				output.writeByteArray(DATA_PAYLOAD_FIELD_NUMBER, (byte[]) originalData, false);
			} else {
				Class<?> dataClass = originalData.getClass();
				Schema dataSchema = RuntimeSchema.getSchema(dataClass);
				try (RecyclableLinkedBuffer rlb = recyclableLinkedBuffer.get()) {
					byte[] array = ProtostuffIOUtil.toByteArray(originalData, dataSchema, rlb.buffer());
					output.writeByteArray(DATA_PAYLOAD_FIELD_NUMBER, array, false);
				}
				output.writeString(DATA_TYPE_FIELD_NUMBER, dataClass.getName(), false);
			}
		}
```

So it will wrote data as is without any additional manipulations (if statement). While in case of GossipRequest data it will go into "else" branch with all manipulations written there since Gossip Request is a Java Class. Obviously first approach much faster. But it will be more relevant to compare deserialization of Message with some Java POJO data (internally inside this POJO you can have byte array btw).

benchmark/src/main/java/test/CacheReadBenchmark.java
antonkharenko: hm, interesting test can you also make a test for failed case where you ask for classes which do not exists.

cluster/src/main/java/io/servicefabric/cluster/gossip/GossipRequest.java
```diff
@@ -7,7 +7,7 @@
 /**
  * Gossip request which be transmitted through the network, contains list of gossips
  */
-final class GossipRequest {
+public final class GossipRequest {
```
antonkharenko: Maybe better to move benchmark tests to the corresponding packages since I am considering Gossip and GossipRequest part of internal implementation, wdyt?

benchmark/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializerBenchmark.java
antonkharenko: Also you put headers into each Gossip so in case of message you have 1xheaders, in case of list10 - 10xheaders and list100 - 100xheaders. Since messages are generally small this headers can significantly increase releavnt sizes of messages. Actually I would say it will be helpful to output actually size of deserailized messages to see (my bet that actual gossip size will be not a 1k, but around 7k-8k).

benchmark/src/main/java/test/CacheReadBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+				return Optional.of(dataClass);
+			} catch (ClassNotFoundException e) {
+				return Optional.absent();
+			}
+		}
+	});
+
+	@Benchmark
+	public void readFromGuavaCache() {
+		for (String arg : GENERIC_ARGS) {
```
antonkharenko: Also you are testing on new strings. It will be interesting to compare how fast they return results of already computed values.

benchmark/src/main/java/test/CacheReadBenchmark.java
kshchepanovskyi: > Here you can see memoizer x10 faster than guava_cache. In MessageSchema we can gain some win if switching it to memoizer.

Also, as I said earlier, you can move to JDK 8.
Then you can use [Map.computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) - it is also faster than Guava cache (but not so powerful).

benchmark/src/main/java/test/CacheReadBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+			"java.lang.Long",
+			"java.lang.Integer",
+			"java.lang.Short",
+			"java.lang.Character",
+			"java.lang.Byte",
+			"io.servicefabric.transport.protocol.MessageSchema",
+	};
+
+	LoadingCache<String, Optional<Class>> guavaCache = CacheBuilder.newBuilder()
+			.expireAfterAccess(1, TimeUnit.HOURS)
```
kshchepanovskyi: Remove `.expireAfterAccess(1, TimeUnit.HOURS)`, otherwise comparison is not fair.

cluster/src/main/java/io/servicefabric/cluster/gossip/GossipRequest.java
artem-v: @antonkharenko good idea, will do

benchmark/src/main/java/test/CacheReadBenchmark.java
artem-v: @antonkharenko accounted 'failed case' in PR#26

benchmark/src/main/java/io/servicefabric/transport/protocol/ProtostuffMessageSerializerBenchmark.java
artem-v: @antonkharenko refactored gossip benchmark test in PR#26 (but still with headers)

benchmark/src/main/java/test/CacheReadBenchmark.java
artem-v: @antonkharenko applied this suggestion in PR#26. Made entire ``CacheReadBenchmark`` with annotation ```@Scope(benchmark)```(instead of ```@Scope(thread)```)    to fully utilize mulltithreading.

artem-v:
Guys, I have created PR#26, let's move discussion there.

Issue #34, opened 2015-09-08
Data type can be lost if message is passed via proxy node (e.g. gossip)


antonkharenko:
Since message doesn't contain field dataType, but only in deserialized form at MessageSchema when message pass via some proxy which can't resolve dataType via class by name (e.g. gossip which pass to some node which doesn't have required data model) then data type will be lost on this node and won't be transferred on next hop.

antonkharenko:
Resolved by PR https://github.com/servicefabric/servicefabric/pull/36

PR #36, opened 2015-09-11
Issue #34 Fix MessageSchema to avoid data type lost

antonkharenko:
if data class can't be resolved by name on intermediate node.

### DIFF
--- a/transport/src/main/java/io/servicefabric/transport/TransportHeaders.java
+++ b/transport/src/main/java/io/servicefabric/transport/TransportHeaders.java
@@ -7,9 +7,22 @@ import rx.functions.Func1;
  */
 public final class TransportHeaders {
 
+  /**
+   * This header is supposed to be used by application in case if same data type can be reused for different message
+   * meanings so it will allow to qualify the specific meaning of data.
+   */
   public static final String QUALIFIER = "q";
 
+  /**
+   * This header is supposed to be used by application in order to correlate request and response messages.
+   */
   public static final String CORRELATION_ID = "cid";
+
+  /**
+   * This is a system header which used by transport for serialization and deserialization purpose. It is not supposed
+   * to be used by application directly and it is subject to changes in future releases.
+   */
+  public static final String DATA_TYPE = "_type";
 
   private TransportHeaders() {
     // Do not instantiate
--- a/transport/src/main/java/io/servicefabric/transport/protocol/MessageSchema.java
+++ b/transport/src/main/java/io/servicefabric/transport/protocol/MessageSchema.java
@@ -3,6 +3,7 @@ package io.servicefabric.transport.protocol;
 import static io.protostuff.LinkedBuffer.MIN_BUFFER_SIZE;
 import static io.servicefabric.transport.utils.RecyclableLinkedBuffer.DEFAULT_MAX_CAPACITY;
 
+import io.servicefabric.transport.TransportHeaders;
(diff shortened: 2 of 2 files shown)


transport/src/main/java/io/servicefabric/transport/TransportHeaders.java
```diff
@@ -11,6 +11,8 @@
 
   public static final String CORRELATION_ID = "cid";
 
+  public static final String DATA_CLASS = "_class";
```
artem-v: Maybe clazz''? )

antonkharenko: it can be even "class" since it is just string key... I was just thinking to mark with underscore kind of internal system fields (for example it is not recommended to start with "_" application level headers)

transport/src/main/java/io/servicefabric/transport/protocol/MessageSchema.java
```diff
@@ -34,15 +35,13 @@
...
-  private static final int DATA_TYPE_FIELD_NUMBER = 4;
+  private static final int DATA_FIELD_NUMBER = 3;
 
   private static final RecyclableLinkedBuffer recyclableLinkedBuffer = new RecyclableLinkedBuffer(MIN_BUFFER_SIZE,
       DEFAULT_MAX_CAPACITY);
 
   private static final Map<String, Integer> fieldMap = ImmutableMap.of("headerKeys", HEADER_KEYS_FIELD_NUMBER,
-      "headerValues", HEADER_VALUES_FIELD_NUMBER, "dataPayload", DATA_PAYLOAD_FIELD_NUMBER, "dataType",
-      DATA_TYPE_FIELD_NUMBER);
+      "headerValues", HEADER_VALUES_FIELD_NUMBER, "dataPayload", DATA_FIELD_NUMBER);
```
artem-v: Here maybe 'data' not 'dataPayload'?

transport/src/main/java/io/servicefabric/transport/protocol/MessageSchema.java
```diff
@@ -135,38 +128,43 @@ public void mergeFrom(Input input, Message message) throws IOException {
...
+    Object data = null;
+    if (dataBytes != null) {
+      String dataClassName = headers.get(TransportHeaders.DATA_CLASS);
+      if (dataClassName == null) {
+        data = dataBytes;
       } else {
-        Optional<Class> optionalDataClass = classCache.get(dataType);
+        Optional<Class> optionalDataClass = classCache.get(dataClassName);
         if (optionalDataClass.isPresent()) {
+          headers.remove(TransportHeaders.DATA_CLASS);
```
artem-v: Sorry, didn't get. Why .remove() is here?

antonkharenko: Since data will be deserialized to java class so we can eliminate this additional header from here.

transport/src/main/java/io/servicefabric/transport/TransportHeaders.java
artem-v: Cool. I like it!

artem-v:
+1. But need test.

antonkharenko:
@artem-v What kind of test do you mean?

artem-v:
@antonkharenko  the short unit test showing that we don't loose dataType header.

PR #45, opened 2015-09-29
Define formatting settings. Part 2

myroslavlisniak:
Part 2 of https://github.com/servicefabric/servicefabric/issues/16
* Add suppression for javadoc check on test
* Used false for join wrapped length property
* Min line count method require javadoc is 4
* Fail build in case of checkstyle violations.

### DIFF
--- a/transport/src/main/java/io/servicefabric/transport/TransportSettings.java
+++ b/transport/src/main/java/io/servicefabric/transport/TransportSettings.java
@@ -3,6 +3,8 @@ package io.servicefabric.transport;
 import javax.annotation.concurrent.Immutable;
 
 /**
+ * Encapsulate transport settings.
+ * 
  * @author Anton Kharenko
  */
 @Immutable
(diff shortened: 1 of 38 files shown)


transport/src/main/java/io/servicefabric/transport/TransportSettings.java
```diff
@@ -101,8 +99,19 @@ public Builder useNetworkEmulator(boolean useNetworkEmulator) {
       return this;
     }
 
+    private int sendHighWaterMark = DEFAULT_SEND_HIGH_WATER_MARK;
+
     public TransportSettings build() {
       return new TransportSettings(this);
     }
+
+    private String logLevel = DEFAULT_LOG_LEVEL;
```
antonkharenko: Why not at the top of a class?

myroslavlisniak: Interesting, it was automatically rearranged.

kshchepanovskyi: If you are using eclipse code formatter, it might be related to latest plugin upgrade.

https://confluence.playtech.corp/display/NL/Formatting

> With latest Eclipse Formatter in IntelliJ, the option "Use Eclipse 4.4 Java formatter" must be enabled!

antonkharenko:
+1

Issue #28, opened 2015-09-08
Improve Cluster join API. Join should be async


myroslavlisniak:
Cluster join API should be async.
Instead of :
```java
ICluster join();
```
It should be something like this: 
```java
ListenableFuture<ICluster> join();
```
or if we don't need to return clutser it can be future of void
```java
ListenableFuture<Void> join();
```

antonkharenko:
In what case returning ICluster in a future may be useful?

myroslavlisniak:
I think returning ICluster is better, because we can write function (callbacks) without capturing of outer variables. Possible use case is send something to the cluster after joining or even start listening  for incoming messages.

myroslavlisniak:
@antonkharenko Can you remind me why we need async join?

antonkharenko:
@myroslavlisniak Actually the main reason is to be consistent with our principles and reactive approach which in a simplified form may be told as "Never block!". Also cluster join is long running operation and it involves several network messages to be sent and received. Also you can always make block on the future so the application calling the join may either block and wait join to be finished or register async callback, but currently there is no choice other then to be blocked on join.

myroslavlisniak:
Merged #46

PR #46, opened 2015-10-13
#28 Improve Cluster join API. Make join async.

myroslavlisniak:
Motivation:
* There should not be blocking calls in Cluster API

Done:
* Make transport start async
* Make cluster start async

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,7 @@
 
         <slf4j.version>1.7.7</slf4j.version>
         <rxjava.version>1.0.6</rxjava.version>
+        <rxjava-guava.version>1.0.3</rxjava-guava.version>
         <guava.version>18.0</guava.version>
         <findbugs.version>2.0.3</findbugs.version>
         <netty.version>4.0.28.Final</netty.version>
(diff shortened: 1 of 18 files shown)


ronenhamias:
+1

transport/src/test/java/io/servicefabric/transport/TransportTest.java
```diff
@@ -624,9 +624,13 @@ private Transport createTransport(TransportEndpoint endpoint, int sendHwm) {
...
             .useNetworkEmulator(true).build());
-    transport.start();
+      try {
+          transport.start().get();
+      } catch (Exception e) {
+          LOGGER.error("Failed to start transport ", e);
+      }
 
-    return transport;
+      return transport;
```
antonkharenko: Why did it shifted? Don't we have correct formatiing (2 spaces) before?

pom.xml
```diff
@@ -94,6 +94,11 @@
                 <version>${rxjava.version}</version>
             </dependency>
             <dependency>
+                <groupId>io.reactivex</groupId>
+                <artifactId>rxjava-guava</artifactId>
+                <version>1.0.3</version>
```
antonkharenko: move version to properties

antonkharenko:
+1

PR #60, opened 2016-07-27
Transport update: setup usage of InetSocketAddress.createUnresolved()

artem-v:
**Updates:**
- no more ``new InetSocketAddress()``  constructor calls; transport code always create unresolved ``InetSocketAddress`` from string ``host`` and ``port``.

- made ``InetSocketAddress``'s inet address resolution happening exactly in **one** place.

- renamed and updated ``ChannelFutureUtils``; made methods more type-friendly, gave sane names.

- new tests in transport to test hostname scenarios.


transport/src/main/java/io/scalecube/transport/TransportEndpoint.java
```diff
@@ -118,43 +112,74 @@ public static InetSocketAddress parseSocketAddress(@CheckForNull String input) {
...
+  }
+
+  /**
+   * Getting local IP4 address by the address of local host. <b>NOTE:</b> that returned IP4 address is expected to be a
+   * publicly visible IP4 address.
+   *
+   * @throws RuntimeException wrapped {@link UnknownHostException} in case when local host name couldn't be resolved
+   *         into an address; or when returned IP4 address belongs to loopback address family.
+   */
+  public static String getLocalIpAddress() {
```
antonkharenko: This method did not cover cases which were covered by IpAddressResolver (e.g. when you have both IPv6 and IPv4 interfaces on your machine). I propose to return back IpAddressResolver and use it.

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -177,19 +181,20 @@ protected void initChannel(Channel channel) {
       }
     });
 
-    ChannelFuture bindFuture = server.bind(localEndpoint.socketAddress());
+    ChannelFuture bindFuture = server.bind(localEndpoint.port());
```
antonkharenko: We don't want to limit on which interface you bind. At least by default you should bind to any interface (0.0.0.0) if it wasn't configured otherwise.

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -241,13 +256,25 @@ public void send(@CheckForNull TransportEndpoint endpoint, @CheckForNull Message
...
-  public void send(@CheckForNull TransportEndpoint endpoint, @CheckForNull Message message,
-      @Nullable SettableFuture<Void> promise) {
+  public void send(@CheckForNull final TransportEndpoint endpoint, @CheckForNull final Message message,
+      @Nullable final SettableFuture<Void> promise) {
     checkArgument(endpoint != null);
     checkArgument(message != null);
-    TransportChannel transportChannel = getOrConnect(endpoint.socketAddress());
     // TODO [AK]: check that channel endpoint id correspond to provided endpoint id; fail otherwise
-    transportChannel.send(message, promise);
+    Futures.addCallback(getOrConnect(endpoint.socketAddress()), new FutureCallback<TransportChannel>() {
```
antonkharenko: Don't register callback on any call. First check that future is completed and if yes, just call get and send; only otherwise register callback.

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -323,29 +350,33 @@ public void onMessage(Message message) {
...
-              connect(channel, address, transportChannel);
-            } else {
-              channel.unsafe().closeForcibly();
-              transportChannel.close();
-            }
+          public TransportChannel compute(final InetSocketAddress input) {
+            final Channel channel = createConnectorChannel();
+            final TransportChannel transportChannel = createConnectorTransportChannel(channel, input);
+
+            Futures.addCallback(registerChannel(channel), new FutureCallback<Void>() {
```
antonkharenko: Lets return back previous version of this code since it was more compact; and basically exactly the same.

transport/src/main/java/io/scalecube/transport/TransportEndpoint.java
```diff
@@ -118,43 +112,74 @@ public static InetSocketAddress parseSocketAddress(@CheckForNull String input) {
...
+    InetAddress inetAddress;
+    try {
+      inetAddress = InetAddress.getLocalHost();
+      if (inetAddress.getAddress()[0] == Byte.MAX_VALUE) {
```
antonkharenko: I don't think this check is relevant

antonkharenko:
+1

PR #69, opened 2016-08-03
Rewrite transport

antonkharenko:


transport/src/main/java/io/scalecube/transport/MessageDeserializer.java
```diff
@@ -1,9 +1,34 @@
...
+import io.netty.buffer.ByteBufInputStream;
+import io.netty.handler.codec.DecoderException;
+import io.protostuff.ProtostuffIOUtil;
+import io.protostuff.Schema;
+import io.protostuff.runtime.RuntimeSchema;
 
-public interface MessageDeserializer {
+public final class MessageDeserializer {
 
-  Message deserialize(ByteBuf bb);
```
artem-v: Move this ``static``  block to ``MessageSchema`` , here just load ``MessageSchema.class`` and init class -- ``Class.forName(MessageSchema.class.getName())``.

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -1,203 +1,161 @@
...
-        @Override
-        public TransportEndpoint apply(TransportHandshakeData handshakeData) {
-          return handshakeData.endpoint();
-        }
-      };
-
   private final TransportEndpoint localEndpoint;
   private final TransportSettings settings;
-  private final EventLoopGroup eventLoop;
-  private final EventExecutorGroup eventExecutor;
```
artem-v: Move 
```java
incomingMessagesSubject = PublishSubject.create();
```
to the ``.start()`` method;   since we already have ``.stop()``  => we can call start, stop, start, stop, start  and ``Transport`` should work fine.

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -1,203 +1,161 @@
...
   @Override
   public TransportEndpoint localEndpoint() {
     return localEndpoint;
   }
 
-  public EventLoopGroup getEventLoop() {
-    return eventLoop;
-  }
-
-  @Override
```
artem-v: Rename ``getEventExecutor()`` to ``getWorkerGroup()``

transport/src/main/java/io/scalecube/transport/TransportSettings.java
```diff
@@ -57,14 +54,31 @@ public boolean isUseNetworkEmulator() {
     return useNetworkEmulator;
   }
```
artem-v: Better to remove ``getWriteBufferHighWaterMark()`` ; otherwise it looks no logical -- we do have HWM and don't have low water mark.

transport/src/main/java/io/scalecube/transport/Transport.java
antonkharenko: Making start/stop logic works requires more considerations. So don't want to include it into this PR.

PR #70, opened 2016-08-06
Remove id from TransportEndpoint and rename it to Address

antonkharenko:


cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetectorSettings.java
```diff
@@ -0,0 +1,72 @@
+package io.scalecube.cluster.fdetector;
+
+public final class FailureDetectorSettings {
```
artem-v: Better name ``FailureDetectorConfig``

cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocolSettings.java
```diff
@@ -0,0 +1,72 @@
+package io.scalecube.cluster.gossip;
+
+public final class GossipProtocolSettings {
```
artem-v: Better name ``GossipProtocolSettings``

transport/src/main/java/io/scalecube/transport/Address.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public final class Address {
```
artem-v: Better name ``TransportAddress`` .  We already have ``ClusterMember`` , i.e. it wasn't not named ``Member``; same is here.

antonkharenko: Address is clear and short name; ClusterMember is a mistake - you are right - I will rename it to just Member.

cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
```diff
@@ -48,35 +48,45 @@
 public final class GossipProtocol implements IGossipProtocol, IManagedGossipProtocol {
   private static final Logger LOGGER = LoggerFactory.getLogger(GossipProtocol.class);
 
-  private ITransport transport;
+  private final String memberId;
+  private final ITransport transport;
+  private final ScheduledExecutorService executor;
```
artem-v: Always wondered why ``gossip`` does have its own executor, while ``clusterMembership`` and ``failureDetector`` don't. Can't we in this PR create dedicated executors for ``clusterMembership`` and ``failureDetector`` -s ?

cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
```diff
@@ -48,35 +48,45 @@
...
+   *
+   * @param memberId id of current member
+   * @param transport transport
+   * @param settings gossip protocol settings
+   */
+  public GossipProtocol(String memberId, ITransport transport, GossipProtocolSettings settings) {
+    this.memberId = memberId;
+    this.transport = transport;
+    this.settings = settings;
+    this.executor = Executors.newSingleThreadScheduledExecutor(createThreadFactory("scalecube-gossip-scheduled-%s"));
```
artem-v: In analogy with ``BootstrapFactory`` this ``"scalecube-gossip-scheduled-%s"`` name must be ``"sc-gossip-%s"``

cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
antonkharenko: It make sense, but there are more issues with executor. I think it deserves separate PR.

PR #72, opened 2016-08-11
Move available port resolving to transport and cosmetic code changes

antonkharenko:


cluster/src/main/java/io/scalecube/cluster/Cluster.java
```diff
@@ -141,19 +100,37 @@ public static ICluster joinAwait(ClusterConfig config) {
...
+        transport = input;
+
+        // Init gossip protocol component
+        gossipProtocol = new GossipProtocol(memberId, transport, config.gossipProtocolConfig);
+
+        // Init failure detector component
+        failureDetector = new FailureDetector(transport, config.failureDetectorConfig);
+
+        // Init cluster membership component
+        clusterMembership = new ClusterMembership(memberId, transport);
```
artem-v: Why ``ClusterMembership`` does not accept ``config`` like  FD and Gossip components?

cluster/src/main/java/io/scalecube/cluster/Cluster.java
```diff
@@ -141,19 +100,37 @@ public static ICluster joinAwait(ClusterConfig config) {
...
+        clusterMembership.setFailureDetector(failureDetector);
+        clusterMembership.setGossipProtocol(gossipProtocol);
+        clusterMembership.setLocalMetadata(config.metadata);
+        clusterMembership.setSeedMembers(config.seedMembers);
```
artem-v: Next setters better to be removed :   ``setSyncTime``, ``setSyncTimeout``, ``setSyncGroup``, ``setMaxSuspectTime``, ``setMaxShutdownTime`` . Use ``config`` instead .

cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
```diff
@@ -309,37 +290,28 @@ private void declareSuspected(Address member) {
...
 
-  private Address selectPingMember(List<Address> members) {
-    if (pingMember != null) {
-      return pingMember;
-    }
-    return members.isEmpty() ? null : selectRandomMembers(members, 1, null).get(0);
+  private Address selectPingMember() {
+    // TODO [AK]: Select ping member not randomly, but sequentially with shuffle after last member selected
+    return !members.isEmpty() ? members.get(ThreadLocalRandom.current().nextInt(members.size())) : null;
   }
```
artem-v: What was wrong with ``selectRandomMembers()`` name and code itself?

cluster/src/main/java/io/scalecube/cluster/gossip/IGossipProtocol.java
```diff
@@ -12,10 +15,19 @@
  */
 public interface IGossipProtocol {
```
artem-v: Following ``start()`` , ``stop()``  methods doesn't need live here, they can be moved to implementation class. Same applies to ``IFailureDetector`` start/stop.

cluster/src/main/java/io/scalecube/cluster/Cluster.java
antonkharenko: It should, just not part of this PR.

cluster/src/main/java/io/scalecube/cluster/Cluster.java
antonkharenko: Also part of Membership refactoring I plan to do.

cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
antonkharenko: select ping member should be actually not random, as I wrote in TODO - in order to have stable upper bound on failure detection time.

And then it is not just selectRandomMembers, but has specific meaning as part of the algorithm (selecting ping request members).

cluster/src/main/java/io/scalecube/cluster/gossip/IGossipProtocol.java
antonkharenko: It depends, if I generally want to have possibility to replace failure detection algorithm with something else - then I need start/stop methods on interface, if not then I don't need interface here at all... not decided here yet.

PR #74, opened 2016-08-12
Covered issue#66: Add new transport settings: listen_address, listen_interface

artem-v:


transport/src/main/java/io/scalecube/transport/TransportConfig.java
```diff
@@ -112,6 +137,21 @@ public String toString() {
...
+      return this;
+    }
+
+    public Builder listenInterface(String listenInterface) {
+      this.listenInterface = listenInterface;
+      return this;
+    }
+
+    public Builder preferIPv6(boolean preferIPv6) {
+      this.preferIPv6 = true;
```
antonkharenko: Always true?

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -335,6 +360,40 @@ public void operationComplete(ChannelFuture channelFuture) throws Exception {
     }
   }
 
+  private static InetAddress getNetworkInterfaceAddress(String intfc, boolean preferIPv6) {
```
antonkharenko: I propose to move logic from this method and from bind0 method to separate class (e.g. IpAddressChooser).

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -141,17 +165,18 @@ public static Transport bindAwait(TransportConfig config) {
     address = Address.createLocal(bindPort);
```
antonkharenko: address here should be listen address!

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -141,17 +165,18 @@ public static Transport bindAwait(TransportConfig config) {
...
 
     ServerBootstrap server = bootstrapFactory.serverBootstrap().childHandler(incomingChannelInitializer);
-    ChannelFuture bindFuture = server.bind(address.host(), address.port());
+    ChannelFuture bindFuture = server.bind(listenAddress, address.port());
```
antonkharenko: here should be address.host() as it was, while address.host() should contain listen address

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -141,17 +165,18 @@ public static Transport bindAwait(TransportConfig config) {
...
     final SettableFuture<Transport> result = SettableFuture.create();
+    final InetAddress finalListenAddress = listenAddress;
```
antonkharenko: why we need this separate finalListenAddress?

transport/src/test/java/io/scalecube/transport/TransportTest.java
```diff
@@ -38,7 +38,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-@SuppressWarnings("unchecked")
+@SuppressWarnings({"unchecked", "ThrowFromFinallyBlock"})
```
antonkharenko: Transport.stop() doesn't throw exception. I think we don't need this annotation at all.

transport/src/main/java/io/scalecube/transport/TransportConfig.java
```diff
@@ -12,6 +12,7 @@
 
   public static final TransportConfig DEFAULT = builder().build();
 
+  public static final boolean DEFAULT_PREFER_IP6 = false;
```
antonkharenko: IPv6

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -131,6 +134,27 @@ public static Transport bindAwait(TransportConfig config) {
    * Starts to accept connections on local address.
    */
   private ListenableFuture<Transport> bind0() {
+    InetAddress listenAddress;
+    if (config.getListenAddress() != null && config.getListenInterface() != null) {
```
antonkharenko: lets check here both null and empty string as valid non set values

transport/src/main/java/io/scalecube/transport/TransportConfig.java
antonkharenko: Lets for consistency add here defaults for listen address/interface too with null (or empty string) values. You can add comment that this means fallback to getLocalHost.

transport/src/main/java/io/scalecube/transport/Addressing.java
```diff
@@ -0,0 +1,210 @@
...
+   * @param port the port to check for availability
+   */
+  static boolean isAvailable(int port) {
+    if (port < MIN_PORT_NUMBER || port > MAX_PORT_NUMBER) {
+      throw new IllegalArgumentException("Invalid port number: " + port);
+    }
+
+    Socket socket = null;
+    try {
+      socket = new Socket(getLocalIpAddress(), port);
```
antonkharenko: hm, once we start to resolve IP address this logic is not really correct anymore. First we need to resolve listen address we are going to bind and then check available port for this specific resolved address. So in other words pass address here and to getNextAvailable{Port}() as one of arguments.

PR #75, opened 2016-08-15
add support for cluster to send using address insted of cluster memeber

ronenhamias:
add support for cluster to send using address insted of cluster memeber


antonkharenko:
I am refactoring cluster membership. So this small will cause unnecessary conflicts. I will embed this change as part of refactoring.

ronenhamias:
ok

PR #76, opened 2016-08-17
Added log4j2 instead logback

artem-v:
Covered issue: https://github.com/scalecube/scalecube/issues/61


transport/pom.xml
```diff
@@ -59,6 +59,16 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
```
antonkharenko: Lets add logger implementation dependencies only to testlib

testlib/src/main/resources/log4j2-test.xml
```diff
@@ -0,0 +1,20 @@
...
+    <Appenders>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout>
+                <pattern>%level{length=1} %date{MMdd-HHmm:ss,SSS} %logger{1.} %message [%thread]%n</pattern>
+            </PatternLayout>
+        </Console>
+    </Appenders>
+
+    <Loggers>
+        <Logger name="io.netty" level="info"/>
```
antonkharenko: Missed - <logger name="io.scalecube.transport" level="info"/> it was not by mistake

transport/pom.xml
artem-v: @antonkharenko in the last commit I have added log4j-test cfg in every module with test code. Dependency on log4j marked as ``optional``.

antonkharenko:
@artem-v This solution looks unnecessary complex. Currently there is only logback.xml and dependency in testlib module and this it. In proposed PR config and dependencies are copy-pasted at several modules (the ones which do not require at all dependency on logger). I tend to reject this PR since there is no difference what logger is used by tests at the moment so I prefer to stay with the simplest possible solution.

artem-v:
@antonkharenko now I put dep on log4j only in testlib.

PR #80, opened 2016-08-26
Update thread model

artem-v:


cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
```diff
@@ -99,16 +93,32 @@ public FailureDetector(Transport transport) {
...
+   * <b>NOTE:</b> this method is for testing purpose only.
+   */
+  ITransport getTransport() {
+    return transport;
+  }
+
+  /**
+   * <b>NOTE:</b> this method is for testing purpose only.
+   */
+  void resetSubject() {
```
antonkharenko: Lets remove it and instead listen for FD events in test until it will check specific member and see what is the result of this check.

antonkharenko:
Changes from this PR migrated to https://github.com/scalecube/scalecube/pull/81

PR #83, opened 2016-09-12
Fixed setting function of FD.members from non-FD thread

artem-v:


cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
```diff
@@ -174,10 +171,8 @@ public void stop() {
     return subject.toSerialized();
   }
 
-  private void doPing() {
-    period++;
-
-    final Address pingMember = selectPingMember();
+  private void doPing(long period, List<Address> members) {
```
antonkharenko: don't pass into this method any parameters

cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
```diff
@@ -193,25 +188,21 @@ private void doPing() {
...
+              LOGGER.trace("No PingAck from {} within {}ms; about to make PingReq now",
+                  pingMember, config.getPingTimeout());
+              doPingReq(pingMember, cid, members);
+            });
 
     transport.send(pingMember, pingMsg);
   }
 
-  private void doPingReq(final Address pingMember, String cid) {
+  private void doPingReq(final Address pingMember, String cid, List<Address> members) {
```
antonkharenko: I propose not to pass here members, but pass members to selectPingReqMembers

cluster/src/main/java/io/scalecube/cluster/fdetector/FailureDetector.java
```diff
@@ -118,15 +115,15 @@ public void setMembers(Collection<Address> members) {
...
+    LOGGER.debug("Updated monitored members[{}]: {}", this.members.size(), this.members);
   }
 
   @Override
   public void start() {
     onPingRequestSubscriber = Subscribers.create(new OnPingRequestAction());
     transport.listen().observeOn(scheduler)
         .filter(PING_FILTER)
-        .filter(targetFilter(transport.address()))
+        .filter(targetFilter())
```
antonkharenko: I think this targetFilter is useless overall. Algorithm can't send Ping not to you. So basically just filtering out this case not valid.

artem-v:
Looking into failed checks

PR #84, opened 2016-09-15
Support leader election

ronenhamias:
Add initial support for leader election algo based on RAFT

this pull request yet assume FD events that might be interesting to the algo


antonkharenko:
Closing it since as I understand there is another PR https://github.com/scalecube/scalecube/pull/85

PR #85, opened 2016-09-16
Scalecube leaderelection

ronenhamias:
support RAFT leader election


antonkharenko:
What is this 49 files?

ronenhamias:
services module is not part of this pull request

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,51 @@
...
+ * expects this heartbeat in case heartbeat does not reach members within X time leader election process 
+ * is initiated. 
+ * 
+ * @author Ronen Nachmias
+ */
+
+public class HeartbeatScheduler {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(HeartbeatScheduler.class);
+  private final ScheduledThreadPoolExecutor scheduler = new ScheduledThreadPoolExecutor(1);
```
myroslavlisniak: Is it really required to have dedicated thread pool for this heartbeat scheduler?
Can we use some shared resource here?

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,51 @@
...
+      public void run() {
+        LOGGER.debug("Leader Node: {} maintain leadership spread {} gossip regards selected leader {}",
+            cluster.address(), RaftProtocol.RAFT_PROTOCOL_HEARTBEAT, cluster.address());
+        cluster.spreadGossip(
+            Message.builder().qualifier(RaftProtocol.RAFT_PROTOCOL_HEARTBEAT).data(cluster.address()).build());
+      }
+    }, heartbeatInterval, heartbeatInterval, TimeUnit.SECONDS);
+  }
+
+  public void stop() {
```
myroslavlisniak: This method is confusing, as for me javadoc here is required for better understanding what it really stops

leaderelection/src/main/java/io/scalecube/leaderelection/MessageListener.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public abstract class MessageListener {
+
+  enum ListenType {
+    GOSSIP, TRANSPORT, GOSSIP_OR_TRANSPORT
+  }
+  
+  public final ICluster cluster;
+
+  public MessageListener(ICluster cluster) {
```
myroslavlisniak: A lot of code duplication in this class.
As I understand  here should be two filter function - "qualifier equal" and "qualifier start with" which pure function and can be defined and saved in the variable and two cases cluster.listen() and cluster.listenGossips() which both return Observable which should be filtered with correct function based on ListenType. Please correct me if I miss something?

leaderelection/src/main/java/io/scalecube/leaderelection/StopWatch.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public abstract class StopWatch {
+
+  private final ScheduledThreadPoolExecutor executor;
+  private final int timeout;
+  private final TimeUnit timeUnit;
+  private final Random rnd;
+
+  public StopWatch(int timeout, TimeUnit timeUnit) {
+    executor = new ScheduledThreadPoolExecutor(1);
```
myroslavlisniak: Same comment as previous, do we need dedicated thread pool here?

leaderelection/src/test/java/io/scalecube/examples/LeaderElectionExample.java
```diff
@@ -0,0 +1,114 @@
...
+ * Created by ronenn on 9/12/2016.
+ */
+
+/*
+ * to run the example use the following commands: start -> this command will initiate seed node join <index> -> this
+ * command will join a node (seed) from the list list -> will list the current cluster state and leaders selected kill
+ * <index> -> will shutdown a node
+ * 
+ */
+public class LeaderElectionExample {
```
myroslavlisniak: Is there any tests for the algorithm except this example?

services/src/test/java/io/scalecube/services/examples/AdSelector.java
```diff
@@ -0,0 +1,18 @@
+package io.scalecube.services.examples;
```
antonkharenko: here is services module

services/src/main/java/io/scalecube/services/state/StateEvent.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * State event emitted by {@code DistributedState} in case of state was changed on one of the cluster members.
+ *
+ * @param <S> State object class. This object should be immutable otherwise state maintenance algorithm won't work
+ *        correctly.
+ * @author Anton Kharenko
+ * @see DistributedState
+ */
+@Immutable
+public final class StateEvent<S> {
```
antonkharenko: here is services module... also many other classes

leaderelection/src/main/java/io/scalecube/leaderelection/MessageListener.java
ronenhamias: the idea is to listen to either gossip or message hence message listener
the idea is to have simple code that does both and nor have the verbose filer function as it will explode the code for each message type

i understand there is code duplication and maybe this can be imoroved

leaderelection/src/test/java/io/scalecube/examples/LeaderElectionExample.java
ronenhamias: not yet will add tests later

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
ronenhamias: you are correct in general
the only question is about canceling the thread as only leaders maintain heartbeats
and i dont want to cancel all other jobs. also this is one thread for the entire node.

any suggestions?

services/src/test/java/io/scalecube/services/examples/AdSelector.java
ronenhamias: wired thing i dont have this in my branch

ronenhamias:
create new clean branch to avoid the services stuff so we can close this one

PR #87, opened 2016-09-18
Support leader election based on RAFT protocol

ronenhamias:
support leader election based on RAFT protocol


antonkharenko:
@ronenhamias This PR misses integration tests (same as we have for GossipProtocolIT, FailureDetectorIT and MembershipIT) which will verify that implementation actually correct and works outside of trivial example. Include also tests for different network issues (message loss and network partitions).

leaderelection/src/main/java/io/scalecube/leaderelection/RaftProtocol.java
```diff
@@ -0,0 +1,10 @@
+package io.scalecube.leaderelection;
+
+/**
+ * Created by ronenn on 9/13/2016.
+ */
+public class RaftProtocol {
+  public static final String RAFT_GOSSIP_HEARTBEAT = "raft.protocol.heartbeat";
```
antonkharenko: Don't create separate empty class for constants, move them into protocol file and also follow qualifier pattern same as for other system messages.

leaderelection/src/test/java/io/scalecube/examples/LeaderElectionExample.java
```diff
@@ -0,0 +1,116 @@
...
+ * 
+ */
+public class LeaderElectionExample {
+
+  public static List<ICluster> nodes = new ArrayList<>();
+  public static Map<ICluster, LeaderElection> leaders = new HashMap<>();
+
+  public static void main(String[] args) {
+
+    System.out.println("to run the example use the following commands: "
```
ronenhamias: this is manual testing example very useful for manual testing i will add simple api based example

leaderelection/src/main/java/io/scalecube/leaderelection/Rules.java
```diff
@@ -0,0 +1,67 @@
...
+import io.scalecube.transport.Message;
+
+import rx.functions.Func1;
+
+/**
+ * Utility class for creating logical filtering rules for message qualifiers.
+ * @author ronenn
+ *
+ */
+public class Rules {
```
ronenhamias: rules was removed

leaderelection/src/test/java/io/scalecube/examples/LeaderElectionExample.java
antonkharenko: ok, add simple example. Also for testing it is pretty convenient to use integration tests.

leaderelection/src/main/java/io/scalecube/leaderelection/MessageListener.java
```diff
@@ -0,0 +1,53 @@
...
+    GOSSIP, TRANSPORT, GOSSIP_OR_TRANSPORT
+  }
+
+  public final ICluster cluster;
+
+  public MessageListener(ICluster cluster) {
+    this.cluster = cluster;
+  }
+
+  void qualifierEquals(final String qualifier, ListenType type) {
```
ronenhamias: message listener was removed

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+  private void killTheLeader(List<LeaderElection> leaders, Address leader) {
+    LeaderElection rm = null;
+    for (LeaderElection m : leaders) {
+      if(leader.equals(m.cluster().address())) {
+        m.cluster().shutdown();
+        rm = m;
+        break;
+      }
+    }
+    leaders.remove(rm);
```
antonkharenko: check here that you actually find some leader

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+
+  
+  @Test
+  public void testLeaderDiesAndNewLeaderIsElected() {
+    ICluster seed = Cluster.joinAwait();
+    LeaderElection seedLeader = createLeaderElection(seed);
+    List<LeaderElection> leaders = createCluster(seed);
+
+    // waiting for a leader to be selected
+    awaitSeconds(30);
```
antonkharenko: Is it really required to wait 30 seconds to select leader?! What is the expect time for the implemented algorithm to choose leader? Please put here as small as possible timeout which is required to choose leader. Also don't put 30 seconds sleep in all other places as well.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+        rm = m;
+        break;
+      }
+    }
+    leaders.remove(rm);
+  }
+
+
+  private List<LeaderElection> createCluster(ICluster seed) {
+    List<LeaderElection> leaders = new ArrayList<>();
```
antonkharenko: this sharing of non thread safe collection between threads is cause for race conditions and missed updates. Also don't create separate executor. You can use joinAwait method or if you want it to be async, then listen for the future returned from Cluster.join method. You can use Guavas helper method https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Futures.html#allAsList(java.lang.Iterable) for utility.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+  @Test
+  public void testAllNodesFollowOneLeader() {
+    ICluster seed = Cluster.joinAwait();
+    LeaderElection seedLeader = createLeaderElection(seed);
+    List<LeaderElection> leaders = createCluster(seed);
+
+    // waiting for a leader to be selected
+    awaitSeconds(30);
+
+    Address seedLeaderAddress = seedLeader.leader();
```
antonkharenko: Is it really so that seed leader always chosen as a leader?

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+        break;
+      }
+    }
+    leaders.remove(rm);
+  }
+
+
+  private List<LeaderElection> createCluster(ICluster seed) {
+    List<LeaderElection> leaders = new ArrayList<>();
+    for (int i = 0 ; i < 50; i++) {
```
antonkharenko: Propose to pass cluster size as parameter so you can test for different cluster sizes. And actually test with different cluster sizes - especially for some corner conditions (e.g. what if 2 members in cluster - will they elect leader and so on?)

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+
+    // waiting for a leader to be selected
+    awaitSeconds(30);
+
+    killTheLeader(leaders,seedLeader.leader());
+    
+    // waiting for a leader to be selected
+    awaitSeconds(30);
+
+    Address newLeader = seedLeader.leader();
```
antonkharenko: You killed seedLeader two lines above, but asking it here for new leader. As I understand seedLeader not actually a leader, so it works. But what if by luck he will be chosen as leader and you will kill it above, so here he won't know about new leader (since he is dead) - this will make test unstable with some probability - so make it stable.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+    
+    for (LeaderElection m : leaders) {
+       Address memberLeaderAddress = m.leader();
+       assertTrue("Expected leader address " + newLeader + ", but actual: " + memberLeaderAddress,
+       memberLeaderAddress.equals(newLeader));
+    }
+    
+    // shutdown everything
+    for (LeaderElection m : leaders) {
```
antonkharenko: Since you need to shutdown everything for each test I propose to move it into helper method.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,118 @@
...
+        public void run() {
+          ICluster member = Cluster.joinAwait(seed.address());
+          leaders.add(createLeaderElection(member));  
+        }
+      });
+    }
+    return leaders;
+  }
+
+
```
antonkharenko: Here required more tests for different adverse conditions. For example:

* Test case when you select leader then disable network from/to this leader, check that in the rest of the cluster new leader elected, also what will happen with this current leader? then enable network againfor this leader and check that this previous leader will agree on new one.

* Split cluster on two partitions one is quorum and another isn't. Check that quorum part will select leader, what will be with non quorum? then join cluster again and check that leader is stable.

* Try to fail some mebers in the process of leader election and check that leader is still elected correctly

* Come up with some more creative test cases

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,62 @@
...
+    String nameFormat = "sc-leaderelection-heartbeat" + cluster.address().toString();
+    scheduler =Executors.newSingleThreadScheduledExecutor(
+        new ThreadFactoryBuilder().setNameFormat(nameFormat).setDaemon(true).build());
+  }
+
+  /**
+   * when becoming a leader the leader schedule heatbeats and maintain leadership followers are expecting this heartbeat
+   * in case the heartbeat will not arrive in X time they will assume no leader.
+   */
+  public void schedule() {
```
antonkharenko: It's pretty hard to read this method's javadoc and also javadoc comment of this class, please fix typos, add punctuation marks, start new sentence from capital letter and so.

leaderelection/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>scalecube-cluster</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<!-- https://mvnrepository.com/artifact/com.unquietcode.tools.jstate/jstate -->
+		<dependency>
+			<groupId>com.unquietcode.tools.jstate</groupId>
```
antonkharenko: The general philosophy is to use as less as possible dependencies. Also regards this specific dependency it seems that it is pretty raw and not really much supported. Also we are using it in a pretty trivial use case, so I propose to code the logic instead of using third party.

ronenhamias: yep i agree just it will be intresting to develop state machine

leaderelection/src/main/java/io/scalecube/leaderelection/LeadershipEvent.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Event which is emitted on cluster leadership changes when member state changed from FOLLOWER, CANDIDATE, LEADER.
+ * 
+ * @author Ronen Nachmias
+ *
+ */
+public class LeadershipEvent {
+
+  private final StateType fromState;
+  private final StateType toState;
```
antonkharenko: I think this type of event not really interesting for public API - it's more like internal state transition. I think that public event should be like, about changes of leader - new leader elected, you get leadership, you lost leadership. Also it's not clear from/to state is related to local member only?  How will I get event about changing in leadership if not local member was elected as leader.

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
```diff
@@ -0,0 +1,357 @@
...
+  private AtomicReference<ScheduledFuture<?>> heartbeatTask;
+  private Subject<LeadershipEvent, LeadershipEvent> subject;
+  
+  /**
+   * Builder of raft leader election.
+   * 
+   * @author Ronen Nachmias
+   *
+   */
+  public static class Builder {
```
antonkharenko: Use config based approach same as it is done for other components.

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,66 @@
...
+    String nameFormat = "sc-leaderelection-heartbeat" + cluster.address().toString();
+    scheduler = Executors.newSingleThreadScheduledExecutor(
+        new ThreadFactoryBuilder().setNameFormat(nameFormat).setDaemon(true).build());
+
+    scheduler.scheduleAtFixedRate(this::spreadHeartbeat, heartbeatInterval, heartbeatInterval, TimeUnit.SECONDS);
```
antonkharenko: I don't really understand why this task was moved to separate task, actually I think it should be part of LeaderElection and also scheduled in the leader election thread.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: no i just used it as reference point for the test propose

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+  }
+
+  private RaftLeaderElection(Builder builder) {
+    checkNotNull(builder.cluster);
+
+    this.cluster = builder.cluster;
+    this.heartbeatInterval = builder.heartbeatInterval;
+    this.leadershipTimeout = builder.leadershipTimeout;
+    this.rnd = new Random();
```
antonkharenko: Don't create new random, but instead use `ThreadLocalRandom.current().nextInt(...)` where needed

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: there is random timeout to select a leader the default is min 15 max 30 seconds (with some random factor)
- when timeout exceeds on first node it start selection
when next node comes leader welcome him specifically so time for his is intimidate
when node dies there are already random clocks waiting for the heartbeats so time should be much less

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
```diff
@@ -0,0 +1,58 @@
...
+    stateMachine.addTransitions(StateType.CANDIDATE, StateType.FOLLOWER);
+    stateMachine.addTransitions(StateType.LEADER, StateType.FOLLOWER);
+
+    stateMachine.addTransitions(StateType.START, StateType.STOPPED);
+    stateMachine.addTransitions(StateType.FOLLOWER, StateType.STOPPED);
+    stateMachine.addTransitions(StateType.CANDIDATE, StateType.STOPPED);
+    stateMachine.addTransitions(StateType.LEADER, StateType.STOPPED);
+
+    stateMachine.onEntering(StateType.FOLLOWER, state -> {
+        becomeFollower();
```
antonkharenko: On what thread this state transition will be executed?

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+import unquietcode.tools.esm.EnumStateMachine;
+
+/**
+ * RAFT state machine dealing with state transition FOLLOWER, CANNIDATE, LEADER.
+ * 
+ * @author Ronen Nachmias
+ */
+public abstract class RaftStateMachine {
```
antonkharenko: Also it's not clear why it's abstract class? Is it supposed to be extended by some other classes except LeaderElection? If not then embed it into leader election.

leaderelection/src/main/java/io/scalecube/leaderelection/LeaderElection.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Created by ronenn on 9/12/2016.
+ */
+public interface LeaderElection {
+
+  Address leader();
+  
+  Observable<LeadershipEvent> listen();
+  
+  ICluster cluster();
```
myroslavlisniak: Do we really need cluster here in the interface?

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
myroslavlisniak: and don't save it to class field

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
```diff
@@ -0,0 +1,357 @@
...
+          leaderSendWelcomeMessage(event.member().address());
+        }
+      });
+
+    transition(StateType.FOLLOWER);
+
+    return this;
+  }
+
+  private void leaderSendWelcomeMessage(Address address) {
```
myroslavlisniak: Looks like address param is not used inside the method

leaderelection/src/test/java/io/scalecube/examples/SimpleLeaderElectionExample.java
```diff
@@ -0,0 +1,34 @@
...
+    });
+    
+    avoidExit();
+  }
+
+  private static void avoidExit() {
+    try {
+      Thread.sleep(60000);
+    } catch (InterruptedException e) {
+      // TODO Auto-generated catch block
```
myroslavlisniak: remove such stuff please

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
myroslavlisniak: I did not get it, so 30 seconds is the worst case leader election expected time with default configuration?

antonkharenko: 30 seconds for test is too long, you should configure it for much smaller numbers on test

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
ronenhamias: follower is in the case:
1. node starts
2. a leader suddenly gets heartbeat for some reason
3. candidate didn't get votes for some time and stuck on candidate state.

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,66 @@
...
+    cluster.spreadGossip(Message.builder()
+        .qualifier(RaftLeaderElection.HEARTBEAT)
+        .data(cluster.address())
+        .build());
+  }
+
+  /**
+   * stop sending heartbeats - it actually means this member is no longer a leader.
+   */
+  public void stop() {
```
ronenhamias: why its different?

it say that it stops to send heartbeat - or am i missing something?

leaderelection/src/main/java/io/scalecube/leaderelection/LeaderElection.java
ronenhamias: i think yes
look on the tests

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
ronenhamias: no - its just that RaftLeaderElection is a StateMachine so it extends StateMachine

leaderelection/src/test/java/io/scalecube/examples/LeaderElectionExample.java
ronenhamias: ok will be in next update

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: ok will be in next update

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: ok will be in addresseed in next update

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: ok will be in addresseed in next update

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
antonkharenko: I am asking on what Thread this state machine library will execute this code?

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
antonkharenko: but still why you need RaftStateMachine and RaftLeaderElection if only RaftLeaderElection can extend it and it's totally useless outside of RaftLeaderElection class.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: CuncurrentLinkedQueue can work as well ?

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
ronenhamias: ok will be in addresseed in next update

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
ronenhamias: leader-election thread
it happens when message trigger observer that heartbeat or vote observer is triggered

leaderelection/src/main/java/io/scalecube/leaderelection/RaftStateMachine.java
ronenhamias: design decesion to remove all the noise from leader election class that handles the actual logic and not the boilerplate

or at least less boilerplate ...

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
ronenhamias: will be addressed in next update.

leaderelection/src/test/java/io/scalecube/examples/SimpleLeaderElectionExample.java
ronenhamias: ok

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
ronenhamias: ok

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
myroslavlisniak: Yeah I miss the point the scheduler executor creates each time when you call schedule method, which is bad and can cause race condition, why just not cancel ScheduledFuture here?

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
```diff
@@ -0,0 +1,66 @@
...
+    this.heartbeatInterval = heartbeatInterval;
+  }
+
+  /**
+   * when becoming a leader the leader schedule heatbeats and maintain leadership followers are expecting this heartbeat
+   * in case the heartbeat will not arrive in X time they will assume no leader.
+   */
+  public void schedule() {
+    String nameFormat = "sc-leaderelection-heartbeat" + cluster.address().toString();
+    scheduler = Executors.newSingleThreadScheduledExecutor(
```
antonkharenko: oh, it's not a constructor, you create a new executor each time when called this method. It should not be like this.

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
antonkharenko: Actually take a look here http://thesecretlivesofdata.com/raft/ It actually says about interval in the range of 150-300 ms, which seems to be pretty fine default values.

leaderelection/src/main/java/io/scalecube/leaderelection/HeartbeatScheduler.java
ronenhamias: i schedule everytime node becomes leader and cancel it once it is not

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: heartbeat every 150 ms???

this is alot of traffic we really need it?

leaderelection/src/main/java/io/scalecube/leaderelection/RaftLeaderElection.java
```diff
@@ -0,0 +1,423 @@
...
+ * with their vote. the candidate starts to collect the votes and check if there is consensus among the cluster members
+ * and that means (N/2 +1) votes once the candidate gained consensus it immediately take leadership and start to gossip
+ * heartbeats to all cluster nodes. a leader node that receives an heartbeat from any other member transition to
+ * follower state in the case where several candidates are running for leadership they both cancel the candidate state
+ * and next timeout node will become a leader and election process is restarted until there is clear consensus among the
+ * cluster https://raft.github.io/
+ * 
+ * @author Ronen Nachmias on 9/11/2016.
+ */
+public class RaftLeaderElection implements LeaderElection {
```
antonkharenko: There is a 'term' property in raft leader election algorithm which is used to resolve leader in case of network partition, I don't see something which would look like this here?

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
```diff
@@ -0,0 +1,126 @@
...
+    Queue<LeaderElection> leaders = new ConcurrentLinkedQueue<LeaderElection>();
+    for (int i = 0; i < numOfNodes; i++) {
+      exec.execute(new Runnable() {
+        @Override
+        public void run() {
+          ICluster member = Cluster.joinAwait(seed.address());
+          leaders.add(createLeaderElection(member));
+        }
+      });
+      awaitSeconds(2);
```
antonkharenko: What the sense to have executor in such case... just start cluster nodes sequentially without any executors

leaderelection/src/test/java/io/scalecube/leaderelection/LeaderElectionIT.java
ronenhamias: yes but this will heartbeat 5-3 messages per second in big clusters its a lot
lets set 500 - 1000 ms

ashwanthkumar:
I just happen to stumble in here randomly. Is there a reason for building RAFT by hand compared to using a well tested library like https://github.com/atomix/copycat ?

antonkharenko:
@ashwanthkumar Hi, thanks for reference, we will review their implementation. Currently RAFT support at ScaleCube is pretty much at research/proof of concept stage.

PR #96, opened 2016-09-22
Remove guava futures usage

myroslavlisniak:
https://github.com/scalecube/scalecube/issues/90

Use CompletableFuture instead of guava's future,
Decrease null usage


cluster/src/main/java/io/scalecube/cluster/Cluster.java
```diff
@@ -178,14 +176,8 @@ public static ICluster joinAwait(ClusterConfig config) {
...
-      });
-
-    return transform(clusterFuture, new Function<Void, ICluster>() {
-      @Override
-      public ICluster apply(@Nullable Void param) {
-        return Cluster.this;
-      }
-    });
+      }, MoreExecutors.directExecutor());
+    return clusterFuture.thenApplyAsync(aVoid -> Cluster.this, MoreExecutors.directExecutor());
```
antonkharenko: I was thinking what if we do it in a more Java 8 like functional style to put instead of `MoreExecutors.directExecutor()` just `Runnable::run`, will it work?

antonkharenko: Also btw, I don't get one thing. Either it make at all any sense to call `thenApplyAsync` and then pass same thread executor. Would not it be better or the same to just call `thenApply`?

cluster/src/main/java/io/scalecube/cluster/membership/MembershipProtocol.java
```diff
@@ -238,13 +236,13 @@ public void stop() {
    * ============== Action Methods ================== *
    * ================================================ */
 
-  private ListenableFuture<Void> doInitialSync() {
+  private CompletableFuture<Void> doInitialSync() {
     LOGGER.debug("Making initial Sync to all seed members: {}", seedMembers);
     if (seedMembers.isEmpty()) {
-      return Futures.immediateFuture(null);
+      return CompletableFuture.completedFuture(null);
```
antonkharenko: hm, actually i see a bug here (not related to futures). We need to call `schedulePeriodicSync();` before returning future.

transport/src/main/java/io/scalecube/transport/ITransport.java
```diff
@@ -49,14 +47,15 @@
 
   /**
    * Sends message to the given address. It will issue connect in case if no transport channel by given {@code address}
-   * exists already. Send is an async operation, if result of operation is not needed leave third
-   * parameter null, otherwise pass {@link SettableFuture}.
+   * exists already. Send is an async operation, if result of operation use {@link ITransport#send(Address, Message)},
```
antonkharenko: you have lost somewhere "is not needed" :)

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -284,55 +284,46 @@ public final void stop(@Nullable SettableFuture<Void> promise) {
...
-          if (channelFuture.isSuccess()) {
-            composeFutures(channelFuture.channel().writeAndFlush(message), promise);
+      channelFuture.addListener((ChannelFuture chFuture) -> {
+          if (chFuture.isSuccess()) {
+            composeFutures(chFuture.channel().writeAndFlush(message), promise);
           } else {
-            if (promise != null) {
-              promise.setException(channelFuture.cause());
-            }
+            promise.completeExceptionally(chFuture.cause());
```
antonkharenko: What will happen if you passed here `COMPLETED_PROMISE` and now try to complete it with exception?

transport/src/test/java/io/scalecube/transport/TransportTest.java
```diff
@@ -161,23 +161,15 @@ public void testPingPongClientTFListenAndServerTFListen() throws Exception {
...
-        assertEquals("Expected clientAddress", client.address(), address);
-        send(server, address, Message.fromQualifier("hi client"));
-      }
+    server.listen().subscribe(message -> {
+      Address address = message.sender();
+      assertEquals("Expected clientAddress", client.address(), address);
+      send(server, address, Message.fromQualifier("hi client"));
     });
 
     // final ValueLatch<Message> latch = new ValueLatch<>();
```
antonkharenko: This comment from stone age can be removed ;-)

cluster/src/test/java/io/scalecube/cluster/fdetector/FailureDetectorIT.java
```diff
@@ -461,4 +459,13 @@ private void assertStatus(
       throw new RuntimeException(e);
     }
   }
+  private <T> CompletableFuture<List<T>> allOf(List<CompletableFuture<T>> futuresList) {
+    CompletableFuture<Void> allFuturesResult =
+            CompletableFuture.allOf(futuresList.toArray(new CompletableFuture[futuresList.size()]));
+    return allFuturesResult.thenApply(v ->
```
myroslavlisniak: Did not found better way how to do it :(
Need to consider better place for this method, because it can be usefull

antonkharenko: Yes, I was thinking about some utility interface with static methods like Futures. But it can be done later, when needed in more places.

transport/src/main/java/io/scalecube/transport/Transport.java
myroslavlisniak: As I understand from spec nothing should happen, this method provides effect only on uncompleted futures

PR #104, opened 2016-11-05
Scalecube service support

antonkharenko:


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,238 @@
...
+
+  private final ICluster cluster;
+
+  private final ServiceRegistry serviceRegistry;
+
+  private final ServiceProxyFactory proxyFactory;
+
+  private final ServiceDispatcher localDispatcher;
+
+  private Microservices(ICluster cluster, Optional<Object[]> services, boolean isSeed) {
```
myroslavlisniak: I don't understand why you need optional here. Usually optional is not used with sequences because you can just use empty sequence. This will simplify code here, in ServiceRegistry

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,238 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(Microservices.class);
+  private static final ServiceProcessor serviceProcessor = new AnnotationServiceProcessor();
+
+  private final ICluster cluster;
+
+  private final ServiceRegistry serviceRegistry;
+
+  private final ServiceProxyFactory proxyFactory;
+
+  private final ServiceDispatcher localDispatcher;
```
myroslavlisniak: This field is unused, I understand that this is start point of listening service messages, but what is the purpose of the field? One use case that I see is shutdown

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -0,0 +1,163 @@
...
+  public Object invoke(Message request, ServiceDefinition definition) throws Exception {
+    Preconditions.checkArgument(definition != null, "Service definition can't be null");
+
+    // Try to call via messaging
+    // Request message
+    if (definition.returnType().equals(CompletableFuture.class)) {
+      if (definition.parametrizedType().equals(Message.class)) {
+        return futureInvokeMessage(request);
+      } else {
+        return futureInvokeGeneric(request);
```
myroslavlisniak: Correct me if I wrong, but looks like these two invocations _futureInvokeMessage_  and _futureInvokeGeneric_ are pretty the same, the only differenece is what should be result of completable future message or message.data. I suggest to avoid such duplication of code, and use some generic method with message transformer function in parameters. e.g futureInvoke(request, message -> message.data())

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -0,0 +1,163 @@
...
+      }
+    } else {
+      CompletableFuture<Object> future = futureInvokeGeneric(request);
+      Object result = future.get();
```
myroslavlisniak: I think this call is dangerous, we should not ever wait forever. Maybe even we should not allow to do sync call, or rethink implementation of it. For example encapsulate method call in function and delay call to this function until we receive result from remote node @antonkharenko, @ronenhamias WDYT?

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -0,0 +1,163 @@
...
+      return result;
+    }
+  }
+
+  
+  private CompletableFuture<Message> futureInvokeMessage(final Message request) throws Exception {
+    final CompletableFuture<Message> messageFuture = new CompletableFuture<>();
+
+    final String correlationId = "rpc-" + UUID.randomUUID().toString();
```
myroslavlisniak: UUID.randomUUID has blocking operation inside, is it necessary here, maybe we can use ThreadLocalRandom and UUID constructor?

services/src/main/java/io/scalecube/services/ServiceProxyFactory.java
```diff
@@ -0,0 +1,94 @@
...
+              return result;
+            }
+          }
+
+        } catch (Throwable ex) {
+          LOGGER.error(
+              "Failed  to invoke service, No reachable member with such service method [{}], args [{}], error [{}]",
+              method, args, ex);
+          if (method.getReturnType().isAssignableFrom(CompletableFuture.class)) {
+            return completeExcptionally(
```
myroslavlisniak: small typo in exceptionally word

services/src/main/java/io/scalecube/services/ServiceRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+        .collect(Collectors.toList());
+  }
+
+  @Override
+  public Optional<ServiceInstance> getLocalInstance(String serviceName) {
+    return Optional.ofNullable(serviceInstances.values().stream()
+        .filter(ServiceInstance::isLocal)
+        .filter(serviceInstance -> serviceInstance.serviceName().equals(serviceName))
+        .findFirst()
+        .orElse(null));
```
myroslavlisniak: findFirst already return optional, you don't need to create it from nullable and put null in orElse

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
myroslavlisniak: Think more about this, I suggest just dont allow sync call at all

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
ronenhamias: +1

services/src/main/java/io/scalecube/services/ServiceRegistry.java
ronenhamias: +1

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
ronenhamias: you are correct and we agreed to totally remove the sync calls

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
ronenhamias: +1

PR #109, opened 2016-11-11
Services support

ronenhamias:
squash all changes


services/src/main/java/io/scalecube/services/IServiceRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface IServiceRegistry {
+
+  void registerService(Object serviceObject);
+
+  void unregisterService(Object serviceObject);
+
+  List<ServiceInstance> serviceLookup(String serviceName);
+
+  ServiceInstance serviceInstance(ServiceReference reference);
```
myroslavlisniak: What will be returned if reference would not be found? Null?
Better than return optional here. Also it would be good to add javadocs for this interface

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,248 @@
...
+
+    /**
+     * Services list to be registered.
+     * 
+     * @param services list of instances decorated with @Service
+     * @return builder.
+     */
+    public Builder services(Object... services) {
+      if (services == null) {
+        this.services = new Object[0];
```
myroslavlisniak: I think this is not good behavior, when services is equal to null it means someone called builder.services(null) which should be forbidden. I believe we should not allow such case

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public ProxyContext router(Class<? extends Router> router) {
+      this.router = router;
+      return this;
+    }
+  }
+
+  private static Map<String, String> metadata(Object... services) {
+    Map<String, String> result = new HashMap<>();
```
myroslavlisniak: You can make this method more functional if you want :)
```java
return Arrays.stream(services).flatMap(service -> {
      Collection<Class<?>> serviceInterfaces = serviceProcessor.extractServiceInterfaces(service);
      return serviceInterfaces.stream().map(serviceProcessor::introspectServiceInterface);
    }).collect(Collectors.toMap(ServiceDefinition::serviceName,def -> "service"));
```

services/src/main/java/io/scalecube/services/ServiceDispatcher.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+    } catch (Exception ex) {
+      repleyWithError(message, ex);
+    }
+
+  }
+
+  private void repleyWithError(Message message, Throwable ex) {
+    Message errorResponseMsg = Message.builder()
+        .data(ex)
```
myroslavlisniak: Do we really need to send exception over network ?

services/src/main/java/io/scalecube/services/ServiceDispatcher.java
```diff
@@ -0,0 +1,102 @@
...
+        .header("exception", "true")
+        .correlationId(message.correlationId())
+        .build();
+    cluster.send(message.sender(), errorResponseMsg);
+  }
+
+  private void handleSync(final ICluster cluster, Message message, Object result) {
```
myroslavlisniak: We agreed to remove sync call, or I missed something?

services/src/main/java/io/scalecube/services/ServiceProxyFactory.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class ServiceProxyFactory {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServiceProxyFactory.class);
+
+  /**
+   * used to complete the request future with timeout exception in case no response comes from service.
+   */
+  private static final ScheduledExecutorService delayer = Executors.newScheduledThreadPool(1);
```
myroslavlisniak: I don't like that we create so much ScheduledExecutor's, we should share resource as much as possible,
Also we should correctly shutdown thread pools or at least make them daemon

services/src/main/java/io/scalecube/services/ServiceRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+        .filter(serviceInstance -> serviceInstance.serviceName().equals(serviceName))
+        .findFirst();
+  }
+
+  public ServiceInstance serviceInstance(ServiceReference reference) {
+    return serviceInstances.get(reference);
+  }
+
+  public Collection<ServiceInstance> services() {
+    return serviceInstances.values();
```
myroslavlisniak: I believe client code can modify this collection after return which is not good

services/src/main/java/io/scalecube/services/annotations/AnnotationServiceProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class AnnotationServiceProcessor implements ServiceProcessor {
+
+  @Override
+  public Collection<Class<?>> extractServiceInterfaces(Object serviceObject) {
+    Class<?>[] interfaces = serviceObject.getClass().getInterfaces();
+    List<Class<?>> serviceInterfaces = new ArrayList<>();
```
myroslavlisniak: More functional variant:
```java
Class<?>[] interfaces = serviceObject.getClass().getInterfaces();
    return Arrays.stream(interfaces)
        .filter(interfaceClass -> interfaceClass.isAnnotationPresent(Service.class))
```

services/src/main/java/io/scalecube/services/annotations/AnnotationServiceProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  public ServiceDefinition introspectServiceInterface(Class<?> serviceInterface) {
+    // Service name
+    Service serviceAnnotation = serviceInterface.getAnnotation(Service.class);
+    checkArgument(serviceAnnotation != null, "Not a service interface: %s", serviceInterface);
+    String serviceName = resolveServiceName(serviceInterface, serviceAnnotation);
+
+    // Method name
+    Map<String, Method> methods = new HashMap<>();
+    for (Method method : serviceInterface.getMethods()) {
```
myroslavlisniak:
```java 
// Method name
    Map<String, Method> methods = Arrays.stream(serviceInterface.getMethods())
        .filter(method -> method.isAnnotationPresent(ServiceMethod.class))
        .collect(Collectors.toMap(method -> {
          ServiceMethod methodAnnotation = method.getAnnotation(ServiceMethod.class);
          return resolveMethodName(method, methodAnnotation);
        }, Function.identity()));
```

services/src/main/java/io/scalecube/services/Microservices.java
ronenhamias: one case if someone call builder services null
but its also possible this node is just a proxy and services not called at all.
i still can gain proxy for the services from this node

you can see in the examples

ronenhamias: but i better add a check there and throw parameter validation exception

services/src/main/java/io/scalecube/services/ServiceDispatcher.java
ronenhamias: we should replay with predefined errors but not part of this phase

services/src/main/java/io/scalecube/services/Microservices.java
myroslavlisniak: if no services at all and was called as services() parameter will be exactly Object[0], and if it was not called at all you have initialized field already

services/src/main/java/io/scalecube/services/ServiceDispatcher.java
myroslavlisniak: I mean do we really need to serialize and send through the network exception, could we send only message and Exception type and recreate it on receiver side, or we need to have remote stacktrace also?

ronenhamias: https://github.com/scalecube/scalecube/issues/110

Issue #93, opened 2016-09-21
MissingBackpressureException and several other exceptions on gossips and membership


antonkharenko:
Under some conditions possible:

```
GossipState{gossip=Gossip{gossipId=897CF2FD84CE291D96EE-5, message=Message {headers: {q=sc/membership/gossip}, sender: null, data: {m: 6AE7216A4A   at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:57)
    at rx.internal.schedulers.ExecutorScheduler$ExecutorSchedulerWorker.run(ExecutorScheduler.java:107)
    at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
    at java.util.concurrent.FutureTask.run(Unknown Source)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(Unknown Source)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
    at java.lang.Thread.run(Unknown Source)
Caused by: rx.exceptions.OnErrorNotImplementedException: PublishSubject: could not emit value due to lack of requests
    at rx.observers.Subscribers$2.onError(Subscribers.java:98)
    at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:152)
    at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:115)
    at rx.internal.operators.OnSubscribeFilter$FilterSubscriber.onError(OnSubscribeFilter.java:90)
    at rx.internal.operators.OperatorObserveOn$ObserveOnSubscriber.checkTerminated(OperatorObserveOn.java:276)
    at rx.internal.operators.OperatorObserveOn$ObserveOnSubscriber.call(OperatorObserveOn.java:219)
    at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:55)
    ... 8 more
Caused by: rx.exceptions.MissingBackpressureException: PublishSubject: could not emit value due to lack of requests
    at rx.subjects.PublishSubject$PublishSubjectProducer.onNext(PublishSubject.java:308)
    at rx.subjects.PublishSubject$PublishSubjectState.onNext(PublishSubject.java:220)
[...]

PR #111, opened 2016-11-13
backpreasure issue #93

ronenhamias:
transport on backpreasure when sending 5000 messages at once
once adding onBackpressureBuffer on listen method it does not fails

### DIFF
--- a/services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
+++ b/services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
@@ -1,5 +1,8 @@
 package io.scalecube.services;
 
+import static io.scalecube.services.ServiceHeaders.service_method_of;
+import static io.scalecube.services.ServiceHeaders.service_request_of;
+
 import static com.google.common.base.Preconditions.checkArgument;
 
 import io.scalecube.cluster.ICluster;
@@ -23,7 +26,7 @@ public class RemoteServiceInstance implements ServiceInstance {
   private final Address address;
   private final String memberId;
   private final String serviceName;
- 
+
 
   /**
    * Remote service instance constructor to initiate instance.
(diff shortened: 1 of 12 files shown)


ronenhamias:
https://github.com/scalecube/scalecube/issues/93

transport/src/main/java/io/scalecube/transport/Transport.java
```diff
@@ -279,7 +279,7 @@ public final void stop(CompletableFuture<Void> promise) {
   @Override
   public final Observable<Message> listen() {
     checkState(!stopped, "Transport is stopped");
-    return incomingMessagesSubject.asObservable();
+    return incomingMessagesSubject.onBackpressureBuffer().asObservable();
```
myroslavlisniak: Can this cause memory overflow? don't we need to limit  buffer?

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -37,6 +37,8 @@ public RemoteServiceInstance(ICluster cluster, ServiceReference serviceReference
     this.cluster = cluster;
     this.address = serviceReference.address();
     this.memberId = serviceReference.memberId();
+    this.futures = new ConcurrentHashMap<>();
+    this.cluster.listen().subscribe(message -> handleReply(message));
```
antonkharenko: 1. Still why do you need to make separate subscription for each separate RemoteServiceInstance and keep separate future map for each service instance. Single future map and subscription would be totally enough for the whole system.

ronenhamias: how this can be done?
the RemoteServiceInstance needs a cluster instance and then subscribe on it
can you please elaborate?

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -78,39 +80,39 @@ private Type extractGenericReturnType(Method method) {
...
-            messageFuture.complete(fn.apply(message));
-          } else {
-            LOGGER.error("cid [{}] remote service invoke respond with error message {}", correlationId, message);
-            messageFuture.completeExceptionally(message.data());
-          }
-        });
 
-    // check that send operation completed successfully else report an error
+    ResponseFuture responseFuture = new ResponseFuture(fn);
+    futures.putIfAbsent(responseFuture.correlationId(), responseFuture);
```
antonkharenko: What about handling methods which return `void`, you will put objects into this map but those methods do not send response so it would be never removed and cause a memory leak. 

Also I don't see where this map of futures is cleared for the case if service response message was lost. It seems that in such case the future will be kept in this map forever so also cause memory leak.

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
antonkharenko: don't subscribe inside remote service instance, but in some separate component.

PR #112, opened 2016-11-19
API Gateway Example

ronenhamias:
API Gateway example demonstrate the basic concept of an API Gateway adapter calling a micro-service the example
creates a micro-cluster consists of 1. gateway node with API Gateway listening on HTTP port 8080 2. GreetingService node
HTTP receive a GET request http://localhost:8080/?name=joe
using a service proxy it calls provider node of GreetingService.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -14,6 +14,20 @@
             <artifactId>scalecube-services</artifactId>
             <version>${project.version}</version>
         </dependency>
+        
+        <!-- https://mvnrepository.com/artifact/org.rapidoid/rapidoid-http -->
+		<dependency>
+			<groupId>org.rapidoid</groupId>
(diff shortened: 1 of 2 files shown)


examples/pom.xml
```diff
@@ -14,6 +14,20 @@
             <artifactId>scalecube-services</artifactId>
             <version>${project.version}</version>
         </dependency>
+        
+        <!-- https://mvnrepository.com/artifact/org.rapidoid/rapidoid-http -->
+		<dependency>
+			<groupId>org.rapidoid</groupId>
+			<artifactId>rapidoid-quick</artifactId>
```
antonkharenko: Don't put such dependency into this repository. If you want to make such kind of high level example, make it in a separate repository

PR #113, opened 2016-11-25
Add Support for service tags for Canary testing

ronenhamias:
when building the cluster so each service can now accept
key/value tags also the example test demo how you can implement a
weighted random service selection for A/B testing of services so when
there are different algo running and gracefully allocating resources to
specific service instance while testing its business performance in the
cluster.
read more about: http://techblog.netflix.com/2013/08/deploying-netflix-api.html

### DIFF
--- a/services/src/main/java/io/scalecube/services/ServiceInfo.java
+++ b/services/src/main/java/io/scalecube/services/ServiceInfo.java
@@ -1,16 +1,10 @@
 package io.scalecube.services;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.databind.ObjectMapper;
-
 /**
- * Helper class used to register service with tags as metadata in the scalecube cluster.
- * parsing from service info to json and back.
+ * Helper class used to register service with tags as metadata in the scalecube cluster. parsing from service info to
+ * json and back.
  */
 public class ServiceInfo {
-
-  @JsonIgnore
-  private static final ObjectMapper mapper = new ObjectMapper();
 
   private String serviceName;
 
@@ -39,41 +33,5 @@ public class ServiceInfo {
 
   private void setTags(Tag[] tags) {
     this.tags = tags;
-  }
-
-  /**
-   * Create service Advertisement from a json format.
-   * @param value json value of a registration info.
-   * @return instance of service Advertisement.
-   */
-  public static ServiceInfo from(String value) {
-    try {
-      return mapper.readValue(value, ServiceInfo.class);
-    } catch (Exception ex) {
-      throw new RuntimeException(ex);
-    }
-  }
-
-  /**
-   * Create json format of a service Advertisement.
-   * @param value instance value of a Advertisement.
-   * @return json value of a registration info.
-   */
-  public static String toJson(ServiceInfo value) {
-    try {
-      return mapper.writeValueAsString(value);
-    } catch (Exception ex) {
-      throw new RuntimeException(ex);
-    }
-  }
-
-  /**
-   * Create json format of a service Advertisement.
-   * @param serviceName given a service name.
-   * @param tags given a service tags.
-   * @return  json value of a registration info.
-   */
-  public static String toJson(String serviceName, Tag[] tags) {
-    return toJson(new ServiceInfo(serviceName, tags));
   }
 }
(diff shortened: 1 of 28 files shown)


services/src/main/java/io/scalecube/services/ServiceInfo.java
```diff
@@ -0,0 +1,79 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Helper class used to register service with tags as metadata in the scalecube cluster.
+ * parsing from service info to json and back.
+ */
+public class ServiceInfo {
+
+  @JsonIgnore
+  private static final ObjectMapper mapper = new ObjectMapper();
```
antonkharenko: Alternative idea for putting tags into metadata without using JSON is to put key-value pairs in specific format.

E.g.:
You can put following property in metadata. You have one property in metadata which defines service name and the you are looking for specific properties in metadata. For example properties which has keys-values formatted as following:
key format: "<service name>:tag:<tag_key>" 
and value equal to the value of tag with key <tag_key>.

So in such way you can reconstruct out of metadata service instances on other nodes without using JSON. 

And btw having this object mapper here is kind of not nice.

services/src/main/java/io/scalecube/services/ServiceRegistryImpl.java
```diff
@@ -115,7 +119,7 @@ private void loadMemberServices(DiscoveryType type, Member member) {
   /**
    * register a service instance at the cluster.
    */
-  public void registerService(Object serviceObject) {
+  public void registerService(Object serviceObject, Tag[] tags) {
```
antonkharenko: Why don't just pass here `Map<String, String> tags` is there any specific benefit for using separate Tag class and pass plain array?

services/src/main/java/io/scalecube/services/ServiceInfo.java
ronenhamias: i was trying to avoid manual parsing inventing format is error prone and a lot of "dirty" code 
as inventing a format is easy... dealing with it is not so nice

services/src/main/java/io/scalecube/services/ServiceRegistryImpl.java
ronenhamias: i think that tag is more clean solution as it describe clean data model - the format of it can be in strings or whatever.

services/src/main/java/io/scalecube/services/ServiceInfo.java
antonkharenko: oh... it didn't show it right... it is not a delimiter:

example is following - you have service with name myCoolService and also you have on this service a tag with name myCoolTag with value 42, then it means that in matadata you are going to have following property:

myCoolService:tag:myCoolTag=42

so since you know that your service name is "myCoolService" and also service names are not allowed to have ":" symbol inside you can safely search over metadata properties for the ones which starts with prefix "myCoolService:tag:", then you remove this prefix to get the value of tag key which is "myCoolTag" and value of property is actually value of the tag.

services/src/main/java/io/scalecube/services/ServiceRegistryImpl.java
antonkharenko: maybe it can be clean if you will use it consistently, but it is used interchangeably in different places with the Map of string so it was difficult for me to follow the logic when review

services/src/main/java/io/scalecube/services/ServiceInfo.java
antonkharenko: Specifically it can be such kind of code:

``` java
  String serviceTagPrefix = "myCoolService:tag:";
  List<Tag> serviceTags = metadata.entrySet().stream()
    .filter(e -> e.getKey().startsWith(serviceTagPrefix))
    .map(e -> new Tag(e.getKey().substring(serviceTagPrefix.length()), e.getValue()))
```

services/src/main/java/io/scalecube/services/ServiceRegistryImpl.java
ronenhamias: i will explain you can see that everywhere its Tag except when reading and writing it to metadata.

PR #114, opened 2016-11-30
Support seperated transport for services

ronenhamias:
make service communication use transport directly and not via cluster to not conflict with cluster management messages


antonkharenko:
Closing this PR since as was discussed it complicates the port dynamic allocation

PR #115, opened 2016-12-03
sometime cluster fails to bind to ports that seem to be taken also when

ronenhamias:
port auto increment is marked as true

this fix do - while and retry to bind again if the root cause is bind
exception and port auto increment is true.

seem that there is race conditions - not always reproduce-able.


antonkharenko:
There are possible race conditions in Transport.bind if you try to start at same time many instances on same machine with same port binding settings. Specifically race condition is possible because several instances can both check for available port and see it is available and then try to bind it. But the fix should not be done in this place and in that way...

ronenhamias:
so where the fix should be done? and what is the proposed solution?

PR #116, opened 2016-12-04
Support service tags

ronenhamias:
support service tag alternative implementation not using json parser and not using Tag model but Map<String,String>

### DIFF
--- a/services/src/main/java/io/scalecube/services/LocalServiceInstance.java
+++ b/services/src/main/java/io/scalecube/services/LocalServiceInstance.java
@@ -2,9 +2,11 @@ package io.scalecube.services;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import io.scalecube.services.ServicesConfig.Builder.ServiceConfig;
 import io.scalecube.transport.Message;
 
 import java.lang.reflect.Method;
+import java.util.Collections;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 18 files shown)


services/src/test/java/io/scalecube/services/a/b/testing/ServiceInfoTests.java
```diff
@@ -0,0 +1,63 @@
...
+  public void test_service_info(){
+    
+    Map<String,String> tags = new HashMap<>();
+    tags.put("key1", "value1");
+    tags.put("key2", "value2");
+    
+    ServiceInfo info = new ServiceInfo("service1", tags);
+    
+    String metadata = info.toMetadata();
+    System.out.println(metadata);
```
antonkharenko: it would be more clear if to put here instead of system.out some assertEquals and check either it equal to expected string.

services/src/main/java/io/scalecube/services/LocalServiceInstance.java
```diff
@@ -82,4 +86,10 @@ public Boolean isLocal() {
   public String toString() {
     return "LocalServiceInstance [serviceObject=" + serviceObject + ", memberId=" + memberId + "]";
   }
+
+
+  @Override
+  public Map<String, String> tags() {
+    return tags;
```
antonkharenko: more safe to `return Collections.unmodifiableMap(tags);`

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -202,12 +200,33 @@ public Builder seeds(Address... seeds) {
...
+    /**
+     * Services list to be registered.
+     * 
+     * @param servicesConfig list of instances decorated with.
+     * @return builder.
+     */
+    public Builder services(ServicesConfig servicesConfig) {
+      checkNotNull(servicesConfig);
+      if (servicesConfig != null) {
+        this.servicesConfig = servicesConfig;
```
antonkharenko: it seems that this method is doing nothing since servicesConfig is inited with `ServicesConfig.empty()` and basically never is null.

PR #117, opened 2016-12-05
Issue#92 Starting many nodes at once throw exception address already in use

antonkharenko:
Fixing following issue https://github.com/scalecube/scalecube/issues/92

### DIFF
--- a/transport/src/test/java/io/scalecube/transport/TransportTest.java
+++ b/transport/src/test/java/io/scalecube/transport/TransportTest.java
@@ -21,12 +21,15 @@ import org.slf4j.LoggerFactory;
 
 import rx.Subscriber;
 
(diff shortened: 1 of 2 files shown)


transport/src/test/java/io/scalecube/transport/TransportTest.java
```diff
@@ -88,6 +90,36 @@ public void testInvalidListenAddress() {
   }
 
   @Test
+  public void testPortAutoIncrementRaceConditions() throws Exception {
```
ronenhamias: maybe should consider to add a test that checks bind exception when port auto increment is false?

antonkharenko: ok, will add

PR #119, opened 2016-12-06
update project status

ronenhamias:
1.0.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Web Site: [http://scalecube.io](http://scalecube.io/)
 
 ## Project Status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ Web Site: [http://scalecube.io](http://scalecube.io/)
 
 ## Project Status
 
-This is now a beta version of the project. 
+This is now version 1.0.2 of the project.
```
antonkharenko: I think we don't need to specify version here... also it will change. We have a maven widget on top of readme page which shows the actual latest version. Also maybe just remove this "Project status" section at all.

PR #120, opened 2016-12-06
Do not create observable filter on each call to cluster's listen method

antonkharenko:


ronenhamias:
the tests are failing

PR #122, opened 2016-12-12
Support service dispatchers

ronenhamias:
Added support for service calls using dispatchers

the service call is a generic way to call service interface using a message while not using a java proxy api.

this enables more generic approach to address services by service-name and service-method-names and not be strongly restricted by a java proxy interface.

it also enables to set timeout per service request (java proxy does not allow this unless you add this as parameter)

### DIFF
--- a/services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
+++ b/services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
@@ -26,6 +26,8 @@ public class RemoteServiceInstance implements ServiceInstance {
   private final String serviceName;
   private final Map<String, String> tags;
 
+  private ServiceRegistry serviceRegistry;
+
 
   /**
    * Remote service instance constructor to initiate instance.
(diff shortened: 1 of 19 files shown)


services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.function.Function;
+
+public class ServiceCall {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServiceProxyFactory.class);
+
+  /**
+   * used to complete the request future with timeout exception in case no response comes from service.
+   */
+  private static final ScheduledExecutorService delayer = Executors.newSingleThreadScheduledExecutor(
```
myroslavlisniak: Again? One more scheduler, I think we have enough of it

services/src/main/java/io/scalecube/services/RemoteServiceInstance.java
```diff
@@ -46,6 +46,29 @@ public String serviceName() {
...
+   * Dispatch a request message and invoke a service by a given service name and method name. expected headers in
+   * request: ServiceHeaders.SERVICE_REQUEST the logical name of the service. ServiceHeaders.METHOD the method name to
+   * invoke.
+   * 
+   * @param request request with given headers.
+   * @return CompletableFuture with dispatching result
+   * @throws Exception in case of an error
+   */
+  public CompletableFuture<Object> dispatch(Message request) throws Exception {
+    ResponseFuture responseFuture = new ResponseFuture(fn -> request);
```
myroslavlisniak: don't like naming: ResponseFuture, DispatchingFuture, is it really futures?

ronenhamias:
done

PR #127, opened 2016-12-21
Improve reliability of GossipProtocol under adverse conditions (message loss 50%)

antonkharenko:
Improvements:
* Do not track infected members and actually allow to resend same gossip from one member to same another member several times (in case of message loss I don't know either first time was delivered and can't mark him as infected).
* Introduced additional gossip config param `gossipFactor` and make to compute max rounds for spread gossips as `gossipFactor * ceilLog2(N)` which seems to be much more reliable settings.
* Select members to spread gossip sequentially and shuffle members once reached end of the list.
* Compute message sent stats and message lost stats at NetworkEmulator and print aggregated stats.

### DIFF
--- a/cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
+++ b/cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
@@ -30,7 +30,6 @@ import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


myroslavlisniak:
Looks ok for me, but can you write a short description about changes in algo which improve reliabilty?

antonkharenko:
@myroslavlisniak Added description of proposed improvements

cluster/src/main/java/io/scalecube/cluster/gossip/GossipProtocol.java
```diff
@@ -176,14 +177,22 @@ private void doSpreadGossip() {
...
     if (gossips.isEmpty()) {
-      return;
+      return; // nothing to spread
     }
 
     try {
+      // Select gossips to send
+      List<Gossip> gossipsToSend = selectGossipsToSend();
+      if (gossipsToSend.isEmpty()) {
+        return; // nothing to spread
```
antonkharenko: need to sweep gossips also in this case as well

myroslavlisniak:
Maybe we can still track infected, but mark member as infected when receive gossip from it, so we definitely know  it is infected? 
@antonkharenko  WDYT?

ronenhamias:
+1

antonkharenko:
@myroslavlisniak experimenting with this idea now... will see either it will give improvements

Issue #132, opened 2016-12-23
Add possibility to change Cluster metadata dynamically after cluster started


antonkharenko:
Currently cluster member metadata can be set only when node started and can't change over the member lifetime.

Add possibility to change metadata for started members. Member is allowed to only his instance metadata. Metadata change is weakly-consistent across the cluster.

antonkharenko:
Fixed at PR: https://github.com/scalecube/scalecube/pull/133

PR #133, opened 2016-12-25
Issue-#132: Add possibility to change Cluster metadata dynamically after cluster started

antonkharenko:


cluster/src/main/java/io/scalecube/cluster/ClusterImpl.java
```diff
@@ -0,0 +1,235 @@
...
+    return messageObservable;
+  }
+
+  @Override
+  public void spreadGossip(Message message) {
+    gossip.spread(message);
+  }
+
+  @Override
+  public Observable<Message> listenGossips() {
```
antonkharenko: no

cluster/src/main/java/io/scalecube/cluster/ClusterImpl.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.concurrent.ConcurrentMap;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Cluster implementation.
+ * 
+ * @author Anton Kharenko
+ */
+final class ClusterImpl implements Cluster {
```
antonkharenko: this PR not related to it. Also I am not sure this is a good idea.

cluster/src/test/java/io/scalecube/cluster/ClusterTest.java
```diff
@@ -29,9 +39,130 @@ public void testJoinDynamicPort() throws Exception {
...
+    for (Cluster node : otherNodes) {
+      node.listenMembership()
+          .filter(MembershipEvent::isUpdated)
+          .subscribe(event -> updateLatch.countDown());
+    }
+
+
+    // Update metadata
+    Map<String, String> updatedMetadata = ImmutableMap.of("key1", "value3");
+    metadataNode.updateMetadata(updatedMetadata);
```
antonkharenko: If you want to update only single property you need to use method updateMetadataProperty. This method will result in full replacement, so if you use updateMetadata then you need to provide all three properties.

ronenhamias: ok

PR #137, opened 2016-12-28
Improve service config

ronenhamias:
- improve service configuration to match with cluster changes
- enables to set cluster config as part of builder
- remove is seed flag
- add travis workaround


pom.xml
```diff
@@ -205,6 +205,10 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.18.1</version>
+                <configuration>
+					<!-- Travis build workaround -->
+					<argLine>-Xms1024m -Xmx2048m</argLine>
+				</configuration>
```
antonkharenko: Formatting issues

PR #144, opened 2016-12-29
add token for yaml

ronenhamias:
add token

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -236,10 +236,6 @@
 				<groupId>org.jacoco</groupId>
 				<artifactId>jacoco-maven-plugin</artifactId>
 				<version>0.7.7.201606060606</version>
-				<configuration>
-		          <destfile>${basedir}/target/coverage-reports/jacoco-unit.exec</destfile>
-		          <datafile>${basedir}/target/coverage-reports/jacoco-unit.exec</datafile>
-		        </configuration>
 				<executions>
 					<execution>
 						<goals>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -66,6 +66,9 @@
         <netty.version>4.0.40.Final</netty.version>
         <protostuff.version>1.4.4</protostuff.version>
         <protobuf.version>2.4.1</protobuf.version>
+        
+    	<argLine>-Xms1024m -Xmx2048m</argLine>
```
antonkharenko: Is it done with some specific purpose to move argLine into properties?

ronenhamias: this is as replacement to configuration since there is some order to the agents that might override it if placed wrong

ronenhamias: here is the log
[INFO] 
[INFO] --- jacoco-maven-plugin:0.7.7.201606060606:prepare-agent (default) @ scalecube-testlib ---
[INFO] argLine set to -javaagent:/home/travis/.m2/repository/org/jacoco/org.jacoco.agent/0.7.7.201606060606/org.jacoco.agent-0.7.7.201606060606-runtime.jar=destfile=/home/travis/build/scalecube/scalecube/testlib/target/jacoco.exec -Xms1024m -Xmx2048m
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ scalecube-testlib ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.

PR #153, opened 2017-01-01
Reduce Tech dept - Code coverage and quality 

ronenhamias:
add example test working with cluster configuration for advanced settings.

 Microservices node = Microservices.builder().clusterConfig(
        ClusterConfig.builder()

            .transportConfig(TransportConfig.builder()
                .port(3000)
                .portAutoIncrement(false)
                .build())) // cluster config
        
        .build(); // node

### DIFF
--- a/services/src/main/java/io/scalecube/services/LocalServiceInstance.java
+++ b/services/src/main/java/io/scalecube/services/LocalServiceInstance.java
@@ -55,11 +55,7 @@ public class LocalServiceInstance implements ServiceInstance {
       if (method.getParameters().length == 0) {
         result = method.invoke(serviceObject);
       } else if (method.getParameters()[0].getType().isAssignableFrom(Message.class)) {
-        if (message.data().getClass().isAssignableFrom(Message.class)) {
-          result = method.invoke(serviceObject, (Message) message.data());
-        } else {
-          result = method.invoke(serviceObject, message);
-        }
+        result = method.invoke(serviceObject, message);
       } else {
         result = method.invoke(serviceObject, new Object[] {message.data()});
       }
--- a/services/src/main/java/io/scalecube/services/ServiceHeaders.java
+++ b/services/src/main/java/io/scalecube/services/ServiceHeaders.java
@@ -34,36 +34,32 @@ public final class ServiceHeaders {
   private ServiceHeaders() {
     // Do not instantiate
   }
-  
+
   /**
-   * Extract header value from a given message.
-   * the service method to invoke on request.
+   * Extract header value from a given message. the service method to invoke on request.
    */
-  public static String service_method_of(Message request) {
+  public static String serviceMethod(Message request) {
     return request.header(METHOD);
   }
 
   /**
-   * Extract header value from a given message.
-   * header is used to mark this registration as a microservice instance.
+   * Extract header value from a given message. header is used to mark this registration as a microservice instance.
    */
-  public static String service_of(Message request) {
+  public static String service(Message request) {
     return request.header(SERVICE);
   }
 
   /**
-   * Extract header value from a given message.
-   * is used to mark a request as a service request.
+   * Extract header value from a given message. is used to mark a request as a service request.
    */
-  public static String service_request_of(Message request) {
+  public static String serviceRequest(Message request) {
     return request.header(SERVICE_REQUEST);
   }
 
   /**
-   * Extract header value from a given message.
-   * header is used to mark a request as a service response.
+   * Extract header value from a given message. header is used to mark a request as a service response.
    */
-  public static String service_response_of(Message request) {
+  public static String serviceResponse(Message request) {
     return request.header(SERVICE_RESPONSE);
   }
 }
--- a/services/src/test/java/io/scalecube/services/DummyRouter.java
+++ b/services/src/test/java/io/scalecube/services/DummyRouter.java
@@ -1,0 +1,22 @@
+package io.scalecube.services;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import io.scalecube.services.routing.Router;
+import io.scalecube.transport.Message;
+
+import java.util.Optional;
+
+public class DummyRouter implements Router {
+  private Object empty;
+
+  public DummyRouter() {
+    checkArgument(this.empty != null);
+  }
+
+  @Override
+  public Optional<ServiceInstance> route(Message request) {
+    return null;
+  }
+
+}
--- a/services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
+++ b/services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
@@ -9,6 +9,11 @@ import java.util.concurrent.TimeUnit;
 final class GreetingServiceImpl implements GreetingService {
 
   @Override
+  public String toString() {
+    return "GreetingServiceImpl []";
+  }
+
+  @Override
   public CompletableFuture<String> greeting(String name) {
     return CompletableFuture.completedFuture(" hello to: " + name);
   }
@@ -31,7 +36,7 @@ final class GreetingServiceImpl implements GreetingService {
   public CompletableFuture<String> greetingNoParams() {
     return CompletableFuture.completedFuture("hello unknown");
   }
-  
+
   @Override
   public CompletableFuture<GreetingResponse> greetingRequest(GreetingRequest request) {
     return CompletableFuture.completedFuture(new GreetingResponse(" hello to: " + request.getName()));
--- a/services/src/test/java/io/scalecube/services/ServiceConfigTest.java
+++ b/services/src/test/java/io/scalecube/services/ServiceConfigTest.java
@@ -1,0 +1,29 @@
+package io.scalecube.services;
+
+import static org.junit.Assert.assertTrue;
+
+import io.scalecube.cluster.ClusterConfig;
+import io.scalecube.transport.TransportConfig;
+
+import org.junit.Test;
+
+public class ServiceConfigTest {
+
+  @Test
+  public void test_services_config_port_increment_false() {
+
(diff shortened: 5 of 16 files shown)


services/src/main/java/io/scalecube/services/ServiceHeaders.java
```diff
@@ -30,40 +30,36 @@
    * header is used to mark a request as a service response.
    */
   public static final String SERVICE_RESPONSE = "service-response";
-
-  private ServiceHeaders() {
-    // Do not instantiate
```
antonkharenko: Why did you remove this constructor it seems you don't want to instantiate this class?

services/src/test/java/io/scalecube/services/DummyRouter.java
```diff
@@ -0,0 +1,23 @@
...
+  private Object empty;
+
+  public DummyRouter() {
+    checkArgument(this.empty != null);
+  }
+
+  @Override
+  public Optional<ServiceInstance> route(Message request) {
+    // TODO Auto-generated method stub
+    return null;
```
antonkharenko: Basically if `null` is supposed behavior then better to return empty optional instead and remove TODO comment above...

services/src/main/java/io/scalecube/services/ServiceHeaders.java
ronenhamias: nope just its all static and code cheker does not like it

services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
```diff
@@ -9,6 +9,11 @@
 final class GreetingServiceImpl implements GreetingService {
 
   @Override
+  public String toString() {
+    return "GreetingServiceImpl []";
```
antonkharenko: Why do you need toString for service? Also default implementation of toString will be about the same informative

services/src/test/java/io/scalecube/services/DummyRouter.java
ronenhamias: this dummy router throw runtime exception when creating it 
this is for covering the case of excpetion when creating router for code coverage

services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
ronenhamias: since i need expected behavior for tests

services/src/test/java/io/scalecube/services/ServiceConfigTest.java
```diff
@@ -0,0 +1,29 @@
...
+        ClusterConfig.builder()
+
+            .transportConfig(TransportConfig.builder()
+                .port(3000)
+                .portAutoIncrement(false)
+                .build())) // cluster config
+        
+        .build(); // node 
+
+    assertTrue(node.cluster().address().port() == 3000);
```
antonkharenko: you need to shutdown node.

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
```diff
@@ -0,0 +1,134 @@
...
+import io.scalecube.transport.Address;
+import io.scalecube.transport.Message;
+
+import org.junit.Test;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Optional;
+
+public class ServiceUnitTest {
```
antonkharenko: ServiceUnitTest seems to be an odd name

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    Message request = Message.builder().correlationId(response.correlationId()).build();
+    Field field = Message.class.getDeclaredField("sender");
+    field.setAccessible(true);
+    field.set(request, member.cluster().address());
+
+    DispatchingFuture dispatcher = DispatchingFuture.from(member.cluster(), request);
+    dispatcher.complete(new Throwable());
+
+    assertTrue(true);
```
antonkharenko: this is seems to be a bit strange check, are you really need it )

services/src/main/java/io/scalecube/services/LocalServiceInstance.java
```diff
@@ -55,11 +55,7 @@ public Object invoke(Message message) throws Exception {
       if (method.getParameters().length == 0) {
         result = method.invoke(serviceObject);
       } else if (method.getParameters()[0].getType().isAssignableFrom(Message.class)) {
-        if (message.data().getClass().isAssignableFrom(Message.class)) {
-          result = method.invoke(serviceObject, (Message) message.data());
-        } else {
-          result = method.invoke(serviceObject, message);
-        }
+        result = method.invoke(serviceObject, (Message) message.data());
```
antonkharenko: Not sure that I know the logic supposed to be here, but is it really that message.data() is always an Message here?

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
```diff
@@ -0,0 +1,134 @@
...
+    assertTrue(instance.toString().equals("LocalServiceInstance [serviceObject=GreetingServiceImpl [], memberId=a]"));
+    assertTrue(instance.tags().isEmpty());
+    assertTrue(instance.memberId().equals("a"));
+    assertTrue(instance.serviceName().equals("b"));
+  }
+
+  @Test
+  public void test_remote_service_instance() {
+
+    Microservices member = Microservices.builder().build();
```
antonkharenko: I think you need to shutdown this member at the end.

services/src/main/java/io/scalecube/services/ServiceHeaders.java
antonkharenko: but the reason to define private default constructor is to forbid anyone to instantiate this class.

services/src/test/java/io/scalecube/services/DummyRouter.java
antonkharenko: still remove TODO comment

services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
antonkharenko: are you testing it somehow via toString? in such case maybe better to test in some more straight forward way.

services/src/main/java/io/scalecube/services/ServiceHeaders.java
ronenhamias: yes but automatic code review does not like it

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
ronenhamias: what is the convention?

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
ronenhamias: this testing dispatcher with throwable

services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
ronenhamias: the test test the to string
that the expected string returns

services/src/main/java/io/scalecube/services/ServiceHeaders.java
antonkharenko: There are no issues related to it.

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
antonkharenko: Typically for unit tests the convention is `<Class name you test>Test`, also I see in this case there are several tests not really related to each other. So maybe create several test classes where each will correspond to specific class you test or give some meaningful name to the test to display the fact why those tests are collocated in single file and remove word `Unit` from the name.

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
antonkharenko: yes, but since the only assert you have here is `assertTrue(true)` which obviously passing no matter what you will do above it is kind of allude that it didn't really test anything.

services/src/test/java/io/scalecube/services/GreetingServiceImpl.java
antonkharenko: Which one is testing this toString?

services/src/test/java/io/scalecube/services/ServiceUnitTest.java
```diff
@@ -0,0 +1,134 @@
...
+    ServiceRegistry registry = new ServiceRegistryImpl(member.cluster(), config, processor);
+    RemoteServiceInstance instance = new RemoteServiceInstance(registry, reference, new HashMap<>());
+
+    assertTrue(instance.toString().equals("RemoteServiceInstance [address=localhost:4000, memberId=a]"));
+    assertTrue(instance.tags().isEmpty());
+    assertTrue(instance.memberId().equals("a"));
+    assertTrue(instance.address().equals(Address.create("localhost", 4000)));
+    assertTrue(!instance.isReachable());
+    assertTrue(!instance.isLocal());
+    assertTrue(instance.serviceName().equals("b"));
```
antonkharenko: I think for such cases it is better to use `assertEquals` instead of `assertTrue` - it will give much more helpful error message in case if assert won't pass.

antonkharenko:
@ronenhamias due to some reasons after your commit it shows many of my comments as outdated discussions even so they wasn't addressed at any way... so please take a look also to outdated discussions since it doesn't make sense to write them second time.

ronenhamias:
sure i will review them

ronenhamias:
@antonkharenko i have related to all your comments in my last commit
i cannot reply anymore in the specific comment due to this outdated stuff

PR #158, opened 2017-01-07
seperate transport for services and use cluster for discovery only

ronenhamias:


services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+    Cluster cluster = Cluster.joinAwait();
+    Transport transport = Transport.bindAwait();
+
+    ServicesConfig services = ServicesConfig.empty();
+    ServiceProcessor serviceProcessor = new AnnotationServiceProcessor();
+    
+    ServiceRegistryImpl registry = new ServiceRegistryImpl(cluster, transport, services, serviceProcessor);
+   
+    try{
+     new ServiceRegistryImpl(null, transport, services, serviceProcessor);
```
antonkharenko: It seems there some formatting missalignment

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+    
+    try{
+      new ServiceRegistryImpl(cluster, transport, null, serviceProcessor);
+     } catch (Exception ex){
+       assertEquals(ex.toString(), "java.lang.IllegalArgumentException: services can't be null");
+     }
+    
+    try{
+      new ServiceRegistryImpl(cluster, transport, services, null);
+     } catch (Exception ex){
```
antonkharenko: Something wrong with formatting

services/src/test/java/io/scalecube/services/ServiceConfigTest.java
```diff
@@ -13,17 +13,22 @@
...
-
+    Microservices node = Microservices.builder()
+        .serviceTransport(TransportConfig.builder()
+            .port(3000)
+            .portAutoIncrement(false)
+            .build())
+        .clusterConfig(ClusterConfig.builder()
             .transportConfig(TransportConfig.builder()
-                .port(3000)
+                .port(4000)
```
antonkharenko: just fyi, actually `ClusterConfig` has shortcut methods for `port` and `portAutoIncrement ` so you can call thm directly there

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -309,4 +334,9 @@ public ProxyContext router(Class<? extends Router> router) {
     return servicesTags;
   }
 
+
+  public Transport transport() {
```
antonkharenko: I just think that this solution is kind of temporary one... so better not to expose via public API transport itself since then later on it will be harder to change stuff if needed, I propose to just expose needed methods there (like send or listen) so then you would have a possibility to internally exchange the implementation without breaking backward compatibility later on.

antonkharenko: Actually I think even further to add into Microservices config boolean property `reuseClusterTransport` so based either this property is set or not it will create new transport or reuse one from cluster.

services/src/main/java/io/scalecube/services/ServiceRegistry.java
```diff
@@ -29,4 +30,6 @@
   
   Cluster cluster();
 
+  Transport transport();
```
antonkharenko: Also I don't think it is totally good idea to expose from this interface Transport, btw the same is true about Cluster also. It seems a little bit awkward that ServiceRegistry interface exposes Cluster/Transport.

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+    try{
+     new ServiceRegistryImpl(null, transport, services, serviceProcessor);
+    } catch (Exception ex){
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: cluster can't be null");
+    }
+    
+    try{
+      new ServiceRegistryImpl(cluster, null, services, serviceProcessor);
+     } catch (Exception ex){
+       assertEquals(ex.toString(), "java.lang.IllegalArgumentException: transport can't be null");
```
antonkharenko: Also it is better to have several separate tests for each of this case, at least it seems for me a bit cleaner.

services/src/main/java/io/scalecube/services/Microservices.java
ronenhamias: +1

cluster/src/test/java/io/scalecube/cluster/ClusterTest.java
```diff
@@ -26,130 +26,144 @@ public void testJoinDynamicPort() throws Exception {
     // Start seed node
     Cluster seedNode = Cluster.joinAwait();
 
-    // Start other nodes
     int membersNum = 10;
-    long start = System.currentTimeMillis();
-    List<Cluster> otherNodes = new ArrayList<>();
-    for (int i = 0; i < membersNum; i++) {
-      otherNodes.add(Cluster.joinAwait(seedNode.address()));
+    List<Cluster> otherNodes = new ArrayList<>(membersNum);
```
antonkharenko: Somethings weird shown here... you should not have diff at that class, since this code is already in the master.

cluster/src/main/java/io/scalecube/cluster/membership/MembershipProtocol.java
```diff
@@ -281,6 +281,8 @@ public void stop() {
         .timeout(config.getSyncTimeout(), TimeUnit.MILLISECONDS, scheduler)
         .subscribe(
             message -> {
+              SyncData syncData = message.data();
```
antonkharenko: Also you should not have this diff as well, since this was added to master already

services/src/test/java/io/scalecube/services/TransportSenderTest.java
```diff
@@ -0,0 +1,65 @@
...
+    TransportSender sender = new TransportSender(transport);
+
+    assertEquals(transport.address(), sender.address());
+
+    CountDownLatch latch = new CountDownLatch(1);
+    sender.listen().subscribe(onNext -> {
+      latch.countDown();
+    });
+
+    latch.await(1, TimeUnit.SECONDS);
```
antonkharenko: all what this line of code is doing is just waiting for 1 second... since you are not sending any message

services/src/test/java/io/scalecube/services/TransportSenderTest.java
```diff
@@ -0,0 +1,67 @@
...
+    sender.send(transport.address(), Message.builder().data("ping").build());
+  }
+
+  @Test
+  public void test_transport_sender_errors() throws InterruptedException {
+
+    try {
+      new TransportSender(null);
+    } catch (Exception ex) {
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: transport can't be null");
```
antonkharenko: maybe it would be simple to test it in a way:

```
@Test(expected= IllegalArgumentException.class)
public void testTransportSenderErrors() {
  new TransportSender(null);
}
```

P.S. Why are you using this php style method naming instead of camel case as it is used everywhere for java?

services/src/test/java/io/scalecube/services/TransportSenderTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    assertEquals(transport.address(), sender.address());
+
+    CountDownLatch latch = new CountDownLatch(1);
+    sender.listen().subscribe(onNext -> {
+      latch.countDown();
+    });
+
+    latch.await(1, TimeUnit.SECONDS);
+    sender.send(transport.address(), Message.builder().data("ping").build());
```
antonkharenko: This won't help since you are waiting for latch before you send the message, also to make any sense out of it, you actually need to check that latch has reached zero...

Also don't forget to shutdown transport at the end of the test (generally I recommend to do it at the finally block)

ronenhamias:
i have merged master again to my branch
so it will not loose track with changes done

services/src/test/java/io/scalecube/services/TransportSenderTest.java
```diff
@@ -0,0 +1,67 @@
...
+      if (error == null) {
+        latch.countDown();
+      }
+    });
+
+    assertTrue(!micro.services().isEmpty());
+    ServiceInstance instance = micro.services().stream().findFirst().get();
+    assertNotEquals(instance.address(), micro.cluster().address());
+    assertEquals(instance.serviceName(), "io.scalecube.services.GreetingService");
+    latch.await(1, TimeUnit.SECONDS);
```
antonkharenko: Here you need to call service/send message and also check that latch actually reached zero.

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: services can't be null");
+    }
+
+    try {
+      new ServiceRegistryImpl(cluster, sender, services, null);
+    } catch (Exception ex) {
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: serviceProcessor can't be null");
+    }
+    
+    cluster.shutdown();
```
antonkharenko: you need to shutdown transport as well

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    Cluster cluster = Cluster.joinAwait();
+    Transport transport = Transport.bindAwait();
+
+    ServicesConfig services = ServicesConfig.empty();
+    ServiceProcessor serviceProcessor = new AnnotationServiceProcessor();
+
+    ServiceCommunicator sender = new ClusterSender(cluster);
+    
+    ServiceRegistryImpl registry = new ServiceRegistryImpl(cluster, sender, services, serviceProcessor);
```
antonkharenko: What this test is checking? I mean is it like the constructor didn't throw any exception? I am not sure that such kind of test is really needed... since if you actually use this class anywhere it obviously will be covered by other test scenarios.

services/src/test/java/io/scalecube/services/TransportSenderTest.java
ronenhamias: @antonkharenko i fixed it to send message and not just wait
not sure why comment is not outdated

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+    try {
+      new ServiceRegistryImpl(cluster, null, services, serviceProcessor);
+    } catch (Exception ex) {
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: transport can't be null");
+    }
+
+    try {
+      new ServiceRegistryImpl(cluster, sender, null, serviceProcessor);
+    } catch (Exception ex) {
+      assertEquals(ex.toString(), "java.lang.IllegalArgumentException: services can't be null");
```
antonkharenko: Maybe better to move each of this case into separate Test and also not really sure that checking exception toString message make much sense, maybe just check the exception is thrown of expected class

services/src/test/java/io/scalecube/services/ServiceConfigTest.java
```diff
@@ -13,17 +13,19 @@
...
-                .build())) // cluster config
-
+    Microservices node = Microservices.builder()
+        .serviceTransport(TransportConfig.builder()
+            .port(3000)
+            .portAutoIncrement(false)
+            .build())
+        .clusterConfig(ClusterConfig.builder()
+            .port(4000)
+            .portAutoIncrement(false)) // cluster config
```
antonkharenko: I am not sure here, but maybe you miss `.build()` call for cluster config?

services/src/test/java/io/scalecube/services/ClusterSenderTest.java
```diff
@@ -0,0 +1,67 @@
...
+    ClusterSender sender = new ClusterSender(cluster);
+
+    assertEquals(cluster.address(), sender.address());
+
+    CountDownLatch latch = new CountDownLatch(1);
+    sender.listen().subscribe(onNext -> {
+      latch.countDown();
+    });
+
+    latch.await(1, TimeUnit.SECONDS);
```
antonkharenko: All the comments which I wrote for `TransportSenderTest` is applicable in same way for this class as well

services/src/test/java/io/scalecube/services/ServiceConfigTest.java
ronenhamias: no not missed it accept builder on purose

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
ronenhamias: transport has shutdown?

services/src/test/java/io/scalecube/services/ServiceRegistryImplTest.java
ronenhamias: it checks that the message make sense

services/src/test/java/io/scalecube/services/ClusterServiceCommunicatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+      if (error == null) {
+        latch.countDown();
+      }
+    });
+
+    assertTrue(!micro.services().isEmpty());
+    ServiceInstance instance = micro.services().stream().findFirst().get();
+    assertEquals(instance.address(), micro.cluster().address());
+    assertEquals(instance.serviceName(), "io.scalecube.services.GreetingService");
+    latch.await(1, TimeUnit.SECONDS);
```
antonkharenko: need to check latch here too

services/src/test/java/io/scalecube/services/ServiceConfigTest.java
antonkharenko: actually this is out of the scope for this PR, but it seems to me that it is a bit unclear in terms of API that you need to pass here non built ClusterConfig and few lines above TransportConfig which is there built... this may be pretty confusing for someone who try to config this stuff. So consider to simplify this approach.

ronenhamias: you are correct just it was not possible before to pass just builder and let the microservices add the additional changes and do the actaul build

better approch is to let the microservices call the build method this way it can apply changes and have defaults

ronenhamias: i suggest to open an issue on this so we can address it later

services/src/test/java/io/scalecube/services/TransportServiceCommunicatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+      if (error == null) {
+        latch.countDown();
+      }
+    });
+
+    assertTrue(!micro.services().isEmpty());
+    ServiceInstance instance = micro.services().stream().findFirst().get();
+    assertNotEquals(instance.address(), micro.cluster().address());
+    assertEquals(instance.serviceName(), "io.scalecube.services.GreetingService");
+    latch.await(1, TimeUnit.SECONDS);
```
antonkharenko: here also

PR #159, opened 2017-01-07
Fix TransportTest.testBlockAndUnblockTraffic

antonkharenko:


ronenhamias:
please check the code coverage issue there is seem to be some degregation

antonkharenko:
@ronenhamias Because the case which I added is not covered by tests, it using first option of address

ronenhamias:
update the branch 
+1

PR #160, opened 2017-01-08
service dependency injection capability 

cloudimpl:


ronenhamias:
you have added +14 missed that are not covered by tests
you can see them here
https://codecov.io/gh/scalecube/scalecube/compare/7a992aeacba4d408993b46ae6325e121fbdece9c...4e13bf39d19cdf31581d7ffebf731df52b13f38d

all the lines marked as red (not red as removed)

cloudimpl:
those path should cover by the latest push. but build was failed in cluster module , before it execute in services module

Issue #171, opened 2017-01-15
Binding transport to IPv6 address causes IllegalFormatConversionException


guss77:
When starting Scalecube and one of the addresses only has an IPv6 address (for various reasons, in my case the DHCP didn't complete, but we could really be using only an IPv6 network) then `Addressing` uses `NetworkInterface.getInetAddresses()` to get an `Inet6Address` which looks something like this: `/fe80:0:0:0:8cf6:f5c8:c946:2c30%eno1`.

Then when `MembershipProtocol` constructor tries to use the address to create a "name format" for `ThreadFactoryBuilder`, the `%` in the address causes the thread factory builder to crash with a `java.util.IllegalFormatConversionException` when it tries to create thread names using `String.format()`.

The full stack trace looks like this:

```
java.util.IllegalFormatConversionException: e != java.lang.Integer
	at java.util.Formatter$FormatSpecifier.failConversion(Formatter.java:4302)
	at java.util.Formatter$FormatSpecifier.printFloat(Formatter.java:2806)
	at java.util.Formatter$FormatSpecifier.print(Formatter.java:2753)
	at java.util.Formatter.format(Formatter.java:2520)
	at java.util.Formatter.format(Formatter.java:2455)
	at java.lang.String.format(String.java:2981)
	at com.google.common.util.concurrent.ThreadFactoryBuilder.format(ThreadFactoryBuilder.java:181)
	at com.google.common.util.concurrent.ThreadFactoryBuilder.setNameFormat(ThreadFactoryBuilder.java:70)
	at io.scalecube.cluster.membership.MembershipProtocol.<init>(MembershipProtocol.java:110)
	at io.scalecube.cluster.ClusterImpl.lambda$join0$25(ClusterImpl.java:81)
	at java.util.concurrent.CompletableFuture.uniCompose(CompletableFuture.java:952)
	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:926)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1962)
	at io.scalecube.transport.TransportImpl.lambda$bind0$0(TransportImpl.java:104)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:514)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:488)
[...]

PR #172, opened 2017-01-15
fix issue #171 Binding transport to IPv6 address causes

ronenhamias:
IllegalFormatConversionException

When starting Scalecube and one of the addresses only has an IPv6
address (for various reasons, in my case the DHCP didn't complete, but
we could really be using only an IPv6 network) then Addressing uses
NetworkInterface.getInetAddresses() to get an Inet6Address which looks
something like this: /fe80:0:0:0:8cf6:f5c8:c946:2c30%eno1.

Then when MembershipProtocol constructor tries to use the address to
create a "name format" for ThreadFactoryBuilder, the % in the address
causes the thread factory builder to crash with a
java.util.IllegalFormatConversionException when it tries to create
thread names using String.format().


ronenhamias:
fix and verify issue: https://github.com/scalecube/scalecube/issues/171

antonkharenko:
A slightly more simple solution: https://github.com/scalecube/scalecube/pull/173

PR #173, opened 2017-01-15
Issue #171: Binding transport to IPv6 address causes

antonkharenko:


ronenhamias:
yep this may work as well :)

PR #176, opened 2017-05-06
add invoke service by service instance (needed by leader election)

ronenhamias:


.travis.yml
```diff
@@ -1,18 +1,17 @@
...
-
-after_success:
-  - bash <(curl -s https://codecov.io/bash) -t a6d36061-7980-4ff6-884b-0311e4d4fc22
-  - ~/jpm/bin/codacy-coverage-reporter -l Java -r ./cluster/target/site/jacoco/jacoco.xml
-  - ~/jpm/bin/codacy-coverage-reporter -l Java -r ./services/target/site/jacoco/jacoco.xml
-  - ~/jpm/bin/codacy-coverage-reporter -l Java -r ./transport/target/site/jacoco/jacoco.xml
\ No newline at end of file
+language: java
+jdk:
+- oraclejdk8
```
antonkharenko: What means ^M ? Is it indicates some issues with line endings?

services/src/test/java/io/scalecube/services/ServiceTest.java
```diff
@@ -1,791 +1,791 @@
...
+import org.junit.Test;
+
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class ServiceTest extends BaseTest {
```
antonkharenko: Why does this class is shown as fully changed even so it doesn't contain any differences?

PR #177, opened 2017-07-03
Service proxy inject

ronenhamias:
**Support for service proxy injection** 

in case service CoarseGrainedServiceImpl  wants to call service GreetingService the builder will inject the service proxy automatically by introspecting the service fields of the local service instances.

this way you can keep your service clean from boiler plate and micro services stuff.

in future should consider to add ability to control the routers of the proxy right now its just using the default round rubin router.

```
public class CoarseGrainedServiceImpl implements CoarseGrainedService {

  @ServiceProxy
  private GreetingService greetingService;

  @Override
  public CompletableFuture<String> callGreeting(String name) {
    return this.greetingService.greeting(name);
  }
}
```


kshchepanovskyi:
Service proxy injection with annotation is handsome, but it is very limited approach:

1. There is no way to customize proxy behavior (like add generic exception/timeout handler, etc).
2. If you use annotation parameters, then you have to apply them in multiple places (DRY!).

I found that following approach is more flexible:

1. Service proxy is a normal spring bean. The most convenient way to create it is to use java config:

```java
@Configuration
class ServicesConfig {
  @Bean
  FooService foo() {
    return proxyFactory.createServiceProxy(FooService.class, options())
  }
}
```

2. You may also intoroduce two qualifier annotations - `@ServiceProxy` and `@LocalService`, if you want to inject proxy or service bean using `@Autowired` (remember, both have same interface). Service and proxy beans should be annotated as `@LocalService` and  `@ServiceProxy` accordingly.

pom.xml
```diff
@@ -1,315 +1,315 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.scalecube</groupId>
+    <artifactId>scalecube-parent</artifactId>
+    <version>1.0.4-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <name>ScaleCube</name>
```
antonkharenko: It seems you didn't make any changes to this file.

services/src/main/java/io/scalecube/services/LocalServiceInstance.java
```diff
@@ -1,104 +1,108 @@
...
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Local service instance invokes the service instance hosted on this local process.
+ * 
+ *
+ */
+public class LocalServiceInstance implements ServiceInstance {
```
antonkharenko: Also some issue with this file since it shown it as fully changed instead of showing actual diffs

ronenhamias: i have fixed the auto CRLF 
git rm --cached -r .
git diff --cached --name-only -z | xargs -0 git add
git commit -m "Fix CRLF"

and its still show them as changed 

any ideas?

ronenhamias:
@kshchepanovskyi thanks for the review few ideas:

when service A wants to call service B i might want to inject:

-  Service Proxy  - java generic proxy with strict contract api. in this case i have timeout configured once for this proxy and cannot control it per request.  

-  Service Dispatcher AKA ServiceCall that is lower level api and gives me more control per invocation but is less strict and more verbose api.  

one more thing i might want to inject is "this" Mircoservices instance, this is useful for when i want to listen on the cluster events sending gossips.


the solution i had in mind is the following:

**using @ServiceProxy**
```
@ServiceProxy()   // default configurations and router
HelloWorld service;
```

```
@ServiceProxy(router=<my router type>.class, timeout=Duration.ofSercounts(1))
HelloWorld service;
```

this is equivalent to:
```
service = microservices.proxy().api(<service api>) 
        .router(<my service router type>)
        .timeout(<duration>)
        .create();
```

**using @ServiceCall**
```
@ServiceCall() // default configuration and router
ServiceCall dispatcher;
```

```
@ServiceCall(router=<my router type>.class, timeout=Duration.ofSercounts(1))
ServiceCall dispatcher;
```

this is equivalent to:
```
service = microservices.dispatcher()
        .router(<my service router type>)
        .timeout(<duration>)
        .create();
```

**using @Microservices**
```
@Microservices()
Microservices microservices;
```

i am not sure about @LocalService annotation since the proxy resolve local services vs remote services automatically.

so currently there is not much configurations to the proxy and service call - its just the type of the endpoint selector / router. 
in this case you are not repeating yourself as you are declaring the type of proxy you want for this specific proxy or service call.

in case you are good with the default behavior you can just not specifiy.
and if you want to change the default behavior we can add the ability to change the default router configuration on the Microservices level:

something like:

'''
Microservices.setDefaultRouter(<my specific router type as global>) // just with less ugly naming convention.
'''

WDYT?

ronenhamias:
@kshchepanovskyi thanks for your review. here is my two cents about invoking / dispatching services:

i am trying to avoid using spring as its overkill and very intrusive to entire application. spring enforce you to build your appication and bootstrap your application in specific way and scalecube has very specific usecases as for where and what you might want to inject.


i was thinking about what as a developer i might want to inject and found these usecases:

.travis.yml
```diff
@@ -1,17 +1,17 @@
...
-  - wget -O ~/codacy-coverage-reporter-assembly-latest.jar $(curl https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/latest | jq -r .assets[0].browser_download_url)
-
-after_success:
-  - bash <(curl -s https://codecov.io/bash) -t a6d36061-7980-4ff6-884b-0311e4d4fc22
-  - java -cp ~/codacy-coverage-reporter-assembly-latest.jar com.codacy.CodacyCoverageReporter -l Java -r ./cluster/target/site/jacoco/jacoco.xml
-  - java -cp ~/codacy-coverage-reporter-assembly-latest.jar com.codacy.CodacyCoverageReporter -l Java -r ./services/target/site/jacoco/jacoco.xml
+language: java
+jdk:
+- oraclejdk8
+notifications:
```
antonkharenko: This file and ServiceTest file still has CRLF issues

PR #178, opened 2017-07-07
@Inject microservices.

ronenhamias:
Enable to inject Microservices instance to the services.
- add Inject annotation for injecting instances that requires no special
params.

- fix the test to inject microservices and see its not null.
- add to service proxy the ability to statically set router and timeout.

```java
 @Inject
  private Microservices ms;

@Inject
  private GreetingService greetingService; // inject with default configuration

// inject with specific configuration.
  @ServiceProxy(router = RoundRobinServiceRouter.class, timeout = 3, timeUnit = TimeUnit.SECONDS)
  private GreetingService greetingService;
```

### DIFF
--- a/services/src/test/java/io/scalecube/services/CoarseGrainedServiceImpl.java
+++ b/services/src/test/java/io/scalecube/services/CoarseGrainedServiceImpl.java
@@ -1,16 +1,27 @@
 package io.scalecube.services;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
+import io.scalecube.services.annotations.Inject;
 import io.scalecube.services.annotations.ServiceProxy;
+import io.scalecube.services.routing.RoundRobinServiceRouter;
 
 public class CoarseGrainedServiceImpl implements CoarseGrainedService {
 
-  @ServiceProxy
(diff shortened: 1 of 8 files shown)


services/src/test/java/io/scalecube/services/CoarseGrainedServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+  @ServiceProxy(router = RoundRobinServiceRouter.class, timeout = 3, timeUnit = TimeUnit.SECONDS)
   private GreetingService greetingService;
 
+  @Inject
+  private Microservices ms;
+
   @Override
   public CompletableFuture<String> callGreeting(String name) {
-    return this.greetingService.greeting(name);
+    if (ms != null) {
```
antonkharenko: I don't really understand the use case. You don't use microservices other than checking it is not null - what is the meaning of it? Maybe it is simpler just not inject `greetingService` if microservices not set and then check that `greetingService` not null.

ronenhamias: you are correct its redundant i have made a commit that clarify this more..

PR #181, opened 2017-07-29
enable the possibility to contribute to metadata and not replace

ronenhamias:
enable the possibility to contribute to metadata and not replace user defined metata when using services

in current solution when using services it will override the cluster
metadata and will put the services metadata on top of the cluster
metadata

this change expose the metadata() accessor from the builder so
micoservices builder will just putAll metadata it has on top of the user
defined.

### DIFF
--- a/services/src/main/java/io/scalecube/services/Microservices.java
+++ b/services/src/main/java/io/scalecube/services/Microservices.java
@@ -33,12 +33,14 @@ import java.util.concurrent.CompletableFuture;
  * compose. True isolation is achieved through shared-nothing design. This means the services in ScaleCube are
  * autonomous, loosely coupled and mobile (location transparent)—necessary requirements for resilence and elasticity
  * 
- * <p>ScaleCube services requires developers only to two simple Annotations declaring a Service but not regards how you
+ * <p>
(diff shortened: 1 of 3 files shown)


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -188,19 +188,20 @@ public Microservices build() {
...
-      Map<String, String> metadata = new HashMap<>();
-
       if (servicesConfig != null && !servicesConfig.services().isEmpty()) {
-        metadata = Microservices.metadata(servicesConfig);
+        Map<String, String> metadata = clusterConfig.metadata();
+        if (metadata.isEmpty()) {
+          metadata = new HashMap<>();
+        }
+        metadata.putAll(Microservices.metadata(servicesConfig));
+        clusterConfig.metadata(metadata);
```
antonkharenko: I propose to change it to:
```
Map<String, String> metadata = new HashMap<>();
metadata.putAll(clusterConfig.metadata());
metadata.putAll(Microservices.metadata(servicesConfig));
clusterConfig.metadata(metadata);
```

ronenhamias: ok

PR #186, opened 2017-08-28
Support rx observable

ronenhamias:
Add support for rx.Observale as return paramter from a service introduces a single request and stream of responses pattern

example:

image

- improve the shotdown process in tests.
- fix service tags issue and renable tests.
-  add example of StockQuote service.
-  remove cluster communicator and always have sepertaed service port.
- improve check arguments and errors.
- add service method to service discovery and fail the request on client side if method is not a service method insted of reaching remote node and than fail.
- make all test extend base test.


transport/src/main/java/io/scalecube/transport/TransportImpl.java
```diff
@@ -44,6 +44,8 @@
 
   private final TransportConfig config;
 
+  
+
```
antonkharenko: remove this change since it only adds clutter

PR #187, opened 2017-08-28
optimize map lookup for correlation id on response

nuwansa:


services/src/main/java/io/scalecube/services/ServiceResponse.java
```diff
@@ -86,7 +86,6 @@ public void complete(final Message message) {
       LOGGER.error("cid [{}] remote service invoke respond with error message {}", correlationId, message);
       this.messageFuture.completeExceptionally(message.data());
     }
-    futures.remove(correlationId);
```
ronenhamias: complete already removes the future
what is the benefit fo calling it explicitly?
response.complete(message); // --> removes 
 this.messageFuture.completeExceptionally(message.data()); // --> removes

nuwansa: handleReply will call before complete for remote response. their we already do one lookup for service response.

Issue #90, opened 2016-09-21
Remove Guava from dependency and use JDK 8 classes instead


antonkharenko:
Currently there is pretty basic usage of Guava, and basically almost all has already analogous at JDK 8. So there is no practical need for additional dependency to Guava.

ronenhamias:
@antonkharenko  i was starting to work on removing guava as dependency and during my work i found that maybe its not so good idea to remove guava just yet. the value of removing guava is very small comparing to the features it brings and from code clarity point of view

there are some obvious things that we might not want to use guava 
but there is also stuff that will just create code duplication of guava.

i suggest to only reduce the dependency on guava for example when i comes to requireNonNull but apart of  that i dont see much value.

so i suggest to close this ticket and maybe open new one with more specific code changes rather than removing guava totally.

antonkharenko:
There is no specific need for this ticket as for now. Idea was that we use guava classes in very trivial places only. But if not it can be kept. Or it can be removed and left where more advanced use cases is needed - like remove from Transport and Cluster and leave in Service or so.

ronenhamias:
services almost not using it mostly cluster and transport use it
the use of services is argument check. that is 3 lines of code.

the more problamatic zones are the InmmutableMaps MultiMaps and thread factory usage in cluster and in transport.

check my pull request and changes it gives some insights

antonkharenko:
ok, I will review them

PR #191, opened 2017-09-23
[WIP] Remove Guava from dependency and use JDK 8 classes instead #90

ronenhamias:
issue #90
- [x] replace checkNotNull with requireNonNull
- [x] replace Strings.isNullOrEmpty
- [x] replace checkArgument
- [ ] replace Throwables.propagate(e);
- [ ] replace immutable maps ans sets


antonkharenko:
@ronenhamias please fix checkstyle & codacy

ronenhamias:
@antonkharenko  i was starting to work on removing guava as dependency and during my work i found that maybe its not so good idea to remove guava just yet
the value of removing guava is very small comparing to the features it brings and from code clarity point of view
there are some obvious that we might not want to use guava
but there is also stuff that will just create code duplication from guava.no
i suggest to only reduce the dependency on guava for example when i comes to requireNonNull but accept that i dont see much value.

so i suggest to close this ticket and maybe open new one with more specific code changes rather than removing guava totally.

antonkharenko:
Lets leave it for now. I will close this PR anyway I think that removing Guava should be done in steps from bottom to top. Like remove it from Transport one PR, remove from Cluster second PR, maybe leave in service if it needed - something like this.

Issue #197, opened 2017-11-13
Scalecube + Docker


mbarker:
I might be going about this wrong, but I'm having some issues with running scale cube within docker. 

So let's say I have 2 docker hosts: `api1` and `api2`. I have a seed service which just initializes the `Cluster`, I'm setting the `port` and using `syncGroup`, and passing in the list of `seeds` (pretty standard stuff). Bake the service into a docker container and run it on `api1` and `api2` passing the seeds of `api1:20000,api2:20000`. 

Each service starts up ok, and then I get a whole bunch of warnings like:
```
Failed to connect from 172.17.0.5:20000 to 172.17.0.8:20000
```

So I wired up some endpoints that spit out some info around the `Cluster` (members). 

The first instance of the service will only have itself as a member, the second instance will have both instances as a member of the cluster (but it still warns every second failing to connect).

I noticed that the ips are all `172.17.x.x` which makes sense, that's what `localhost` resolves to for each of the docker containers. I figured that was the problem (they aren't on the same network), so I saw I could set what address/interface gossip should bind to. I tried several combinations without luck (using hostname or local ip of the docker hosts throws an exception, failing to bind within the docker container because it's not part of the network interfaces). Without bridging the docker networks between docker hosts, is there a way I can set this up to work between docker hosts, or from a docker host to a non docker host within the same network?

ronenhamias:
Hello and thanks for your comment.

First:
did you have a look on the microservices project that runs the scalecube as dockers?
https://github.com/scalecube/microservices

can you describe little more about the topology you have is it on same host or across different hosts and docker engines?

one thing that comes to my mind is to create a network between hosts using SWARM maybe?

we can discuss it on 
https://gitter.im/scalecube/Lobby

mbarker:
@ronenhamias I know we walked through a few options in our discussion, but this is sort of what I had in mind (to avoid changing our deployment strategy, or build a custom network across docker hosts).
#198

antonkharenko:
@mbarker It seems to me that more ordinary approach would be to create network using [docker network](https://docs.docker.com/engine/reference/commandline/network/) commands and then start docker containers on that network. Also, we have created separate small tool to do it even from Java code at https://github.com/scalecube/docker-utils, check readme for some examples on this specific part. 

[...]

PR #198, opened 2017-11-15
Optional host address

mbarker:
Related to #197

### DIFF
--- a/cluster/src/main/java/io/scalecube/cluster/membership/MembershipProtocolImpl.java
+++ b/cluster/src/main/java/io/scalecube/cluster/membership/MembershipProtocolImpl.java
@@ -102,10 +102,10 @@ public final class MembershipProtocolImpl implements MembershipProtocol {
   public MembershipProtocolImpl(Transport transport, MembershipConfig config) {
     this.transport = transport;
     this.config = config;
-    Member member = new Member(IdGenerator.generateId(), transport.hostAddress(), config.getMetadata());
+    Member member = new Member(IdGenerator.generateId(), transport.address(), config.getMetadata());
     this.memberRef = new AtomicReference<>(member);
 
-    String nameFormat = "sc-membership-" + Integer.toString(transport.hostAddress().port());
+    String nameFormat = "sc-membership-" + Integer.toString(transport.address().port());
     this.executor = Executors.newSingleThreadScheduledExecutor(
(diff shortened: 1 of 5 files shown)


antonkharenko:
@mbarker Generally, I would like to elaborate a little more on your use case at https://github.com/scalecube/scalecube/issues/197 before merging this pull request. But anyway I think it would be reasonable to change it the following way:

* In TransportConfig what is currently called listenAddress/Port rename to bindAddress/Port
* In TransportConfig what you have introduced as hostAddress rename it to actually listenAddress and add listenPort since in general case it also may be different (e.g. you can map it on different port using docker -p option)
* The method Transport.address() should actually return listenAddress:Port (if not set then bindAddress:Port) so in such case you don't need to make any changes in MembershipProtocol.

mbarker:
@antonkharenko Yep that seems pretty reasonable. I was hesitant to replace `Transport.address()` since that was used in lots of other places, and I was trying to limit the scope of the change to only publishing a different address/port. I think it totally makes sense to have a `bindAddress`/`bindPort` and `listenAddress`/`listenPort`. That certainly would work for our use case.

I discussed some other options with @ronenhamias on how I could use scalecube under docker. His suggestions all boiled down to all of the containers need to be on the same network (can be achieved with docker swarm/kubernetes/custom network shared across docker hosts). The reason I made the PR was I wanted to avoid changing the way we planning on deploying our containers. Also I wanted a better explanation of what I was thinking.

I can definitely provide more details about our use case if you'd like. Let me know what you want to know.

mbarker:
I should also mention part of our use case is having docker containers and non docker containers (standalone java processes) form a `Cluster`.

ronenhamias:
i guess my suggestion to use swarm or such is just one possible option that fits specific usecase and its not directly related to the specific change suggested as in this pull request. 
as i understand the possibility to select the host address to listen on is legit requirement regardless.

i was just wondering if listenAddress is not doing the same? maybe the name is not that intuitive and as anton suggested should be named bindAddress

ronenhamias:
@antonkharenko @mbarker i am trying to understand the change and few questions came to my mind:

its not clear whats the difference between  Address and HostAddress:
Address - Returns local Address on which current instance of transport listens for incoming messages.
HostAddress - Returns host Address on which the current instance is accessible for incoming messages

also @antonkharenko :
what is the difference between bindAddress and listenAddress?

mbarker:
Closing in favor of #203

PR #217, opened 2018-03-02
ExchangeStream for Service-to-Service messaging

ronenhamias:


ipc/src/main/java/io/scalecube/ipc/ChannelContext.java
```diff
@@ -71,59 +69,56 @@ public Address getAddress() {
...
   }
 
   public void postReadError(Throwable throwable) {
-    eventSubject.onNext(new Event.Builder(ReadError, this).error(throwable).build());
+    subject.onNext(new Event.Builder(ReadError, this).error(throwable).build());
   }
 
   public void postMessageWrite(ServiceMessage message) {
-    eventSubject.onNext(new Event.Builder(MessageWrite, this).message(message).build());
+    subject.onNext(new Event.Builder(MessageWrite, this).message(message).build());
```
ronenhamias: this function postMessageWrite looks exactlly like postReadSuccess
do we really need 2?

PR #270, opened 2018-03-27
Develop stream processor unsubscribe

ronenhamias:


ronenhamias:
moved this work to streams

PR #286, opened 2018-04-18
measure real latency on response received in SimpleStressTest (WIP)

snripa:


ronenhamias (commented):
build fails with styling

PR #307, opened 2018-08-19
Added cleaning resources

segabriel:


artem-v (commented):
Plz add unit tests. It looks like that aint' hard to do because you depend only on intterfaces

rsocket-transport/src/main/java/io/scalecube/services/transport/rsocket/RSocketServiceTransport.java
```diff
@@ -67,8 +67,12 @@ public EventLoopGroup getExecutorService() {
   }
 
   @Override
-  public Mono<Void> shutdown(ExecutorService executorService) {
-    return Mono.defer(
-        () -> FutureMono.from((Future) ((EventLoopGroup) executorService).shutdownGracefully()));
+  public Mono shutdown(ExecutorService executorService) {
+    if (executorService == null) {
```
dmytro-lazebnyi: Wrap whole method to Mono.defer?

services-api/src/main/java/io/scalecube/services/transport/api/ServiceTransport.java
```diff
@@ -46,5 +46,5 @@ static ServiceTransport getTransport() {
    * @param executorService transport executor service
    * @return shutdown signal
    */
-  Mono<Void> shutdown(ExecutorService executorService);
+  Mono shutdown(ExecutorService executorService);
```
dmytro-lazebnyi: Why `Void` is removed? What are we expecting from return type `Mono`?

services-discovery/src/main/java/io/scalecube/services/discovery/ScalecubeServiceDiscovery.java
```diff
@@ -86,8 +86,10 @@ public ServiceEndpoint endpoint() {
 
   @Override
   public Mono<Void> shutdown() {
-    sink.complete();
-    return Mono.fromFuture(cluster.shutdown());
+    return Mono.defer(() -> {
+      sink.complete();
+      return cluster != null ? Mono.fromFuture(cluster.shutdown()) : Mono.empty();
```
dmytro-lazebnyi: In accordance with similar changes maybe better to use `Optional`?

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -322,7 +326,12 @@ private GatewayBootstrap addConfig(GatewayConfig config) {
     }
 
     private Mono<Void> shutdown() {
-      return Flux.fromIterable(gatewayInstances.values()).flatMap(Gateway::stop).then();
+      return Mono.defer(
+          () ->
+              gatewayInstances != null && !gatewayInstances.isEmpty()
```
dmytro-lazebnyi, line 331: In accordance with similar changes maybe better to use `Optional`?

services-api/src/main/java/io/scalecube/services/transport/api/ServiceTransport.java
artem-v: @dmytro-lazebnyi good point, I will set it back

services-discovery/src/main/java/io/scalecube/services/discovery/ScalecubeServiceDiscovery.java
artem-v: @dmytro-lazebnyi yep

services/src/main/java/io/scalecube/services/Microservices.java
artem-v: @dmytro-lazebnyi nope

PR #308, opened 2018-08-19
Support service registry events

ronenhamias:


services-api/src/main/java/io/scalecube/services/registry/api/RegistryEvent.java
```diff
@@ -0,0 +1,44 @@
...
+public class RegistryEvent {
+
+  public enum Type {
+    ADDED, REMOVED;
+  }
+
+  private ServiceReference serviceReference;
+  private Type type;
+
+  public RegistryEvent(Type type, ServiceReference serviceReference) {
```
artem-v: Make it private plz.

artem-v: Make a shortcut factory methods: `createAdded()` and `createRemoved()` , sort of. Not neccessarly with that names.

services-api/src/main/java/io/scalecube/services/registry/api/RegistryEvent.java
```diff
@@ -0,0 +1,44 @@
...
+    this.serviceReference = serviceReference;
+    this.type = type;
+  }
+
+  public RegistryEvent(RegistryEvent e) {
```
artem-v: Make it private plz.

services-api/src/main/java/io/scalecube/services/registry/api/ServiceRegistry.java
```diff
@@ -20,4 +21,6 @@
   boolean registerService(ServiceEndpoint serviceEndpoint);
 
   ServiceEndpoint unregisterService(String endpointId);
+
```
artem-v, line 25: Add `stop` or better `close()` . We need to have cleanup phase since we do have `listen()`  hence we have a subscribers layer hence we have to have a cleanup phase to tell subscribres bye bye.

artem-v (approved):

artem-v (changes requested):
Need to have a cleanup

services-api/src/main/java/io/scalecube/services/registry/api/ServiceRegistry.java
ronenhamias: done

artem-v (approved):

Issue #331, opened 2018-09-08
Subscribing and processing of ServiceReference's events


segabriel:
First of all, what is the [ServiceReference](https://github.com/scalecube/scalecube-services/blob/d4971485ac8d5aa0bcf9449c887b3756365ebd8a/services-api/src/main/java/io/scalecube/services/ServiceReference.java#L14)? It is a flat structure of each method in each service in the specified endpoint.

I have two different cases in the exchange project.

1. I want to know when some qualifier will be available in the cluster to invoke (subscribe to) it. And it doesn't matter how many microservices will be available, at least 1. 
Ok. Right now I have two approaches to achieve it: 
```java
Microservices ms = ...
GreetingService greetingService = ms.call().create().api(GreetingService.class);

ms.discovery().listen()
  .filter(DiscoveryEvent::isRegistered)
  .flatMap(discoveryEvent -> 
    Flux.fromIterable(discoveryEvent.serviceEndpoint().serviceRegistrations()))
  .filter(serviceRegistration ->
    serviceRegistration.namespace().equals(GreetingService.SERVICE_NAME))
  .take(1)
  .subscribe($ -> greetingService.invoke().subscribe())
```
and the next one (note it will be work correctly after merging #330):
```java
Microservices ms = ...
GreetingService greetingService = ms.call().create().api(GreetingService.class);

ms.serviceRegistry().listen()
  .filter(RegistryEvent::isAdded)
  .filter(registryEvent -> 
    registryEvent.serviceReference().namespace().equals(GreetingService.SERVICE_NAME))
  .take(1)
  .subscribe($ -> greetingService.invoke().subscribe())
```
So these are very similar approaches, I think we need to leave only one of them because they look like a duplicate logic (two publishers does the same). And we shouldn't reveal [the discovery](https://github.com/scalecube/scalecube-services/blob/d4971485ac8d5aa0bcf9449c887b3756365ebd8a/services/src/main/java/io/scalecube/services/Microservices.java#L379) instance at all, because it has a lot of important methods with side effects like `start` and `shutdown`.

[...]

PR #342, opened 2018-09-12
Simplified service discovery and refactoring of registry events

segabriel:
part of #331

**Motivation**:
For now, we have two approaches to subscribe to an available/unavailable endpoint events. But the way via discovery has a side effect. because we can receiver `registered event` with some endpoint inside it and this endpoint is not registered in the service registry yet.


services-api/src/main/java/io/scalecube/services/registry/api/ServiceRegistry.java
```diff
@@ -23,7 +23,9 @@
 
   ServiceEndpoint unregisterService(String endpointId);
 
-  Flux<RegistryEvent> listen();
+  Flux<ReferenceRegistryEvent> listenReferenceEvents();
```
ronenhamias: i dont like this contract its bloated to have this kind in the interface

maybe we can provide more clean interface?

services-api/src/main/java/io/scalecube/services/registry/api/ServiceRegistry.java
```diff
@@ -23,7 +23,9 @@
...
+
+  Flux<EndpointRegistryEvent> listenEndpointEvents();
```
ronenhamias: same as above

services-api/src/main/java/io/scalecube/services/registry/api/ServiceRegistry.java
ronenhamias: something like:

```
listen(RegistryEvent::reference)

listen(RegistryEvent::endpoint)
```

artem-v (approved):

PR #346, opened 2018-09-13
Added member host and port to discovery config and fixed cluster codec

segabriel:


artem-v (approved):
Check locally; docker image is being built

PR #354, opened 2018-09-15
Changed setting discovery options

segabriel:
**Motivation:**
Recently I faced with the unexpected behavior of the Microservices' builder. I wanted to specify a member host and a seed address to my instance. 
```java
    Microservices.builder()
        .seeds(seeds)
        .discoveryConfig(
            DiscoveryConfig.builder()
                .memberHost(memberHost))
...
```
And this construction doesn't merge the specified props into the one config. It has lost the given seeds.
I think it will be better to make this more clear than it's now. And I'd like to propose to remove redundant setters (`seed`, `discoveryPort`) and use only one and this will look like:
```java
        Microservices.builder()
            .discovery(options -> options
                .seeds(seeds)
                .memberHost(memberHost))
...
```
it is a more clear approach.


artem-v (approved):
+1

PR #368, opened 2018-10-07
Defer address lookup to subscription 

omerglam:
When executing a remote call i expect nothing to happen until i subscribe on the publisher (as per [reactor reference](http://projectreactor.io/docs/core/release/reference/#reactive.subscribe) )

### DIFF
--- a/services/src/main/java/io/scalecube/services/ServiceCall.java
+++ b/services/src/main/java/io/scalecube/services/ServiceCall.java
@@ -322,14 +322,17 @@ public class ServiceCall {
   }
 
   private Mono<Address> addressLookup(ServiceMessage request) {
-    return Mono.fromCallable(() -> router.route(serviceRegistry, request)
-      .map(ServiceReference::address)
-      .orElseThrow(() -> noReachableMemberException(request)))
-      .doOnError(
-        t -> {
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -322,15 +322,14 @@ public ServiceCall create() {
...
-              Optional<Object> dataOptional = Optional.ofNullable(request.data());
-              dataOptional.ifPresent(ReferenceCountUtil::safestRelease);
-            });
+    return Mono.fromCallable(() -> router.route(serviceRegistry, request)
+      .map(ServiceReference::address)
+      .orElseThrow(() -> noReachableMemberException(request)))
+      .doOnError(
+        t -> {
+          Optional<Object> dataOptional = Optional.ofNullable(request.data());
+          dataOptional.ifPresent(ReferenceCountUtil::safestRelease);
```
artem-v: Would you mind if I ask you to inline `dataOptional` and geti rid of curly braces block.

artem-v (changes requested):
Great PR. Minor improvement is required.

artem-v (approved):

PR #387, opened 2018-10-23
Updated cluster version

segabriel:


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -113,7 +113,7 @@
   private final Consumer<ServiceDiscoveryConfig.Builder> discoveryOptions;
 
   private Microservices(Builder builder) {
-    this.id = IdGenerator.generateId();
+    this.id = UUID.randomUUID().toString();
```
dmytro-lazebnyi, line 116: What is the motivation for these changes?

segabriel: we removed utils-module from the current version of the cluster and right now I don't want to add a dependency only for id generation.

pom.xml
```diff
@@ -50,11 +50,17 @@
...
+      <dependency>
+        <groupId>io.scalecube</groupId>
+        <artifactId>scalecube-cluster</artifactId>
+        <version>${scalecube-cluster.version}</version>
+      </dependency>
       <dependency>
-        <groupId>${project.groupId}</groupId>
-        <artifactId>scalecube-utils</artifactId>
+        <groupId>io.scalecube</groupId>
+        <artifactId>scalecube-transport</artifactId>
```
artem-v: Remove plz. Why it's here?

services-api/src/main/java/io/scalecube/services/transport/api/Address.java
```diff
@@ -0,0 +1,47 @@
+package io.scalecube.services.transport.api;
```
artem-v: Move it to services-api  pzl. Let's not add additional dependencies on transport where it was absent.

services-discovery/pom.xml
```diff
@@ -14,7 +14,10 @@
     <dependency>
       <groupId>io.scalecube</groupId>
       <artifactId>scalecube-cluster</artifactId>
-      <version>${scalecube-cluster.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.scalecube</groupId>
+      <artifactId>scalecube-transport</artifactId>
```
artem-v: Remove plz

pom.xml
segabriel: it's the main pom and these are in `dependencyManagement`

services-discovery/src/main/java/io/scalecube/services/discovery/ScalecubeServiceDiscovery.java
```diff
@@ -103,7 +105,10 @@ public ServiceEndpoint endpoint() {
   private ClusterConfig.Builder clusterConfigBuilder(ServiceDiscoveryConfig config) {
     Builder builder = ClusterConfig.builder();
     if (config.seeds() != null) {
-      builder.seedMembers(config.seeds());
+      builder.seedMembers(
+          Arrays.stream(config.seeds())
+              .map(address -> Address.create(address.host(), address.port()))
```
artem-v: Would u mind if I ask u to fully specify Address type here. Just to be in sync with line 48

artem-v (approved):

pom.xml
segabriel: however, yes, we can, the cluster depends on this transport

Issue #404, opened 2018-10-26
Provide shorter and nicer api for setting up a service transport


artem-v:
Microservices on line 251 has such a nice api for configuring discovery config: 
```java
    public Builder discovery(Consumer<ServiceDiscoveryConfig.Builder> discoveryOptions) {
      this.discoveryOptions = discoveryOptions;
      return this;
    }
```
And inherently in ServiceDiscoveryConfig:
```java
  public static Builder builder(Consumer<Builder> discoveryOptions) {
    Builder builder = new Builder();
    if (discoveryOptions != null) {
      discoveryOptions.accept(builder);
    }
    return builder;
  }
```
Make similar api for setting up service transport. At the moment service transport builder methods just scattered on Microservices class.

artem-v:
This is done.

PR #406, opened 2018-10-27
Provide shorter and nicer api for setting up a service transport #404

snripa:


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -253,23 +295,8 @@ public Builder discovery(Consumer<ServiceDiscoveryConfig.Builder> discoveryOptio
...
-    }
-
-    public Builder servicePort(int port) {
-      this.transportBootstrap.servicePort(port);
-      return this;
-    }
-
-    public Builder numOfThreads(int numOfThreads) {
-      this.transportBootstrap.numOfThreads(numOfThreads);
+    public Builder transportConfig(ServiceTransportConfig serviceTransportOptions) {
```
segabriel: it should look like the following
```java
public Builder transport(Consumer<ServiceTransportConfig.Builder> transportOptions) {
   ...
}
```

services/src/main/java/io/scalecube/services/transport/ServiceTransportConfig.java
```diff
@@ -0,0 +1,93 @@
...
+import io.scalecube.services.transport.api.ServiceTransport;
+import io.scalecube.services.transport.api.WorkerThreadChooser;
+
+public class ServiceTransportConfig {
+  private Integer servicePort;
+  private String serviceHost;
+  private ServiceTransport transport;
+  private Integer numOfThreads;
+  private WorkerThreadChooser workerThreadChooser;
+
```
artem-v, line 13: Fields are missing `final` qualifier

services/src/main/java/io/scalecube/services/Microservices.java
artem-v: 100%

artem-v (approved):

artem-v (dismissed):
Pressed _Approve_ by mistake last time

dmytro-lazebnyi (approved):

Issue #405, opened 2018-10-27
Change Gateway api service provider


artem-v:
Consider to change (and much simpify life for clients).

From :
```java
interface Gateway {
  Mono<InetSocketAddress> start(...);
  Mono<Void> stop()
}
```
think to change to : 
```java
interface Gateway {
  Mono<Gateway> start(...);
  InetSocketAddress address();
  Mono<Void> stop()
}
```
Rationale: 
- in first variant u have to map gateway object instance somewhere in order to  later to call .stop on it.
- in second variant u have much more freedom u are not  coupled with address, but the an address is still being provided.

artem-v:
This is done.

PR #407, opened 2018-10-27
Change Gateway api service provider #405

snripa:


examples/src/main/java/io/scalecube/examples/gateway/HttpStubGateway.java
```diff
@@ -12,24 +12,32 @@
 
 public class HttpStubGateway implements Gateway {
 
+  private InetSocketAddress address;
```
artem-v, line 15: Put a `final` here and create two constructors, this way you would not have such blocks of code as ` .map(tick -> this)`  and assume that `this ` already contains what we want. Better build logic around immutable thing.

services/src/main/java/io/scalecube/services/gateway/Gateway.java
```diff
@@ -31,13 +31,20 @@ static Gateway getGateway(Class<? extends Gateway> gatewayClass) {
...
       GatewayConfig config,
       Executor workerThreadPool,
       boolean preferNative,
       Call call,
       Metrics metrics);
 
+  /**
+   * Returns Gateway's address if it's started, {@code null} otherwise.
+   *
+   * @return Mono of listening address of gateway if it's started, otherwise - empty.
```
artem-v: > otherwise - empty

Nope. Never null. We just shouldn't give GW a chance to start if it can't bind.

artem-v (approved):

artem-v (dismissed):
Pressed _Approve_ by mistake last time

examples/src/main/java/io/scalecube/examples/gateway/HttpStubGateway.java
segabriel: it's an example

services/src/main/java/io/scalecube/services/gateway/Gateway.java
segabriel: done

segabriel (approved):

artem-v (approved):

PR #450, opened 2018-11-23
Send error on empty data service call

### LINKED ISSUE #445: Don't filter invalid response -> turn it into error response
In `ServiceCall` at lines 357 and 365 there's a `.filter()` operator: 
```java
  private static Function<? super Mono<ServiceMessage>, ? extends Publisher<ServiceMessage>> asMono(
      boolean isRequestTypeServiceMessage) {
    return mono ->
        isRequestTypeServiceMessage
            ? mono
            : mono.filter(ServiceMessage::hasData).map(ServiceMessage::data);
  }
```
1) Turn this block of code to simple `biFuntion (Bool, Mono) -> Publisher`  or `(Bool, Flux) -> Publisher `. 
2) Remove filtering by having msgs only with data when `isRequestTypeServiceMessage` is `false`. If return type is not a msgs and doesn't contain data then this is error at the level of service contract or service api layer; probably service returns null somehow.

snripa:
Closes #445


snripa:
@artem-v are you sure we want these changes exactly as described in issue or in some other places as well?

services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -36,56 +37,69 @@
...
-     *     form this {@link Call}
-     */
-    public Call(
-        ClientTransport transport,
-        ServiceMethodRegistry methodRegistry,
-        ServiceRegistry serviceRegistry) {
-      this.transport = transport;
-      this.serviceRegistry = serviceRegistry;
-      this.methodRegistry = methodRegistry;
+  private static ServiceMessage toServiceMessage(MethodInfo methodInfo, Object... params) {
```
segabriel: Please remove `static` modifier and move at the end of the class, and the same for `toServiceMessage`, `noReachableMemberException`, `toStringOrEqualsOrHashCode`. See `addressLookup`.

services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -36,56 +37,69 @@
   private final ServiceRegistry serviceRegistry;
   private final Router router;
 
+  private final Function<ServiceMessage, Object> msgToResp =
+      sm -> sm.hasData() ? sm.data() : EMPTY_RESPONSE_ERROR;
+
+  private final Function<Boolean, Function<Mono<ServiceMessage>, Mono<Object>>> monoRespFn =
```
segabriel: IMHO, it was better as a method `asMono` and `asFlux`.

snripa: that's not the same and previous ones would have been also uglier if we added `EMPTY_RESPONSE_ERROR` instead of filtering empty ones

segabriel: ok, I got it, why do we need to use for it some fields instead of methods?

snripa: there is no difference

services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -288,7 +252,7 @@ public ServiceCall create() {
               Optional<Object> check =
                   toStringOrEqualsOrHashCode(method.getName(), serviceInterface, params);
               if (check.isPresent()) {
-                return check.get(); // toString, hashCode was invoked.
+                return check.get(); // toString, hashCode was been invoked.
```
artem-v: : ) There's no such thing as `was been`,  u meant `has been` ?

services/src/main/java/io/scalecube/services/ServiceCall.java
artem-v: Actually, it's better to refactor existing `asMono` and `asFlux`  instead of having exacly same thing but done via some tricky field definition. Why somebody has to go over field  and then dive into some logic. Fields are fields. Agree with @segabriel . What we need INHO is just get rid of that silent damn `.filter(ServiceMessage::hasData)`  and that's it. Possibly with minimal changes.

services/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -36,59 +37,22 @@
   private final ServiceRegistry serviceRegistry;
   private final Router router;
 
+  private final Function<ServiceMessage, Object> msgToResp =
```
artem-v: Move it to `private` function.

services/src/main/java/io/scalecube/services/ServiceErrors.java
```diff
@@ -0,0 +1,14 @@
+package io.scalecube.services;
+
+import io.scalecube.services.api.ErrorData;
+import io.scalecube.services.api.Qualifier;
+import io.scalecube.services.api.ServiceMessage;
+
+public interface ServiceErrors {
+
+  ServiceMessage EMPTY_RESPONSE_ERROR =
```
artem-v: Let's not have separate class for this, just def it as part of logic of a function as proposed in comment at [line 41](https://github.com/scalecube/scalecube-services/pull/450/files#diff-c402059b5663b50c40754f48b7677f9cR40)

artem-v (approved):

artem-v:
> @artem-v are you sure we want these changes exactly as described in issue or in some other places as well?

Im 100% sure we shouldn't silently drop a message response

services/src/main/java/io/scalecube/services/ServiceCall.java
ronenhamias: has being

snripa: damn it i read it "has been")

services/src/main/java/io/scalecube/services/ServiceErrors.java
snripa: That could be a good place to collect different types of errors... ok, let's create it as we have the second error to collect there

segabriel (approved):
:+1:

services/src/main/java/io/scalecube/services/ServiceCall.java
snripa: done. Though, that still looks not so good because of `mono.cast(Objcet.class)`... 
For now I don't know how to avoid that(

artem-v:
Would you mind to fix conflicts?

PR #454, opened 2018-11-27
Added rsocket aeron services transport

segabriel:


ronenhamias (dismissed):
+1

ronenhamias:
1. we should not yet merge aeron transport as our default transport rather just offer it as alternative to netty.
2/ we should be able to run both transports side by side provisioning service endpoints both over aeron and netty at the same time.

artem-v:
Agreed to close.

PR #458, opened 2018-12-02
Micrometer metrics

snripa:
The initial implementation of using Micrometer metrics. 
Csv reporter added as it's used for perf. tests metrics reporting.


services/src/main/java/io/scalecube/services/metrics/CsvMeterRegistry.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.stream.StreamSupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Implementation of {@link io.micrometer.core.instrument.MeterRegistry} that publishes metrics to
+ * csv files (one per metric) periodically.
+ */
+public class CsvMeterRegistry extends StepMeterRegistry {
```
ronenhamias, line 41: this CSV thing should be as part of this project?

services-benchmarks/src/main/java/io/scalecube/services/benchmarks/gateway/distributed/DistributedBenchmarkState.java
```diff
@@ -38,7 +39,7 @@ protected void beforeAll() throws Exception {
             .gateway(GatewayConfig.builder("rsws", RSocketGateway.class).build())
             .gateway(GatewayConfig.builder("ws", WebsocketGateway.class).build())
             .gateway(GatewayConfig.builder("http", HttpGateway.class).build())
-            .metrics(registry())
+            .metrics(null)
```
dmytro-lazebnyi, line 42: Should `null` be here?

PR #459, opened 2018-12-03
Update version of reactor netty

aharonha:
Use the same version that [cluster](https://github.com/scalecube/scalecube-cluster/blob/develop/pom.xml#L27)


segabriel (approved):
But please investigate which version is used in rsocket (https://github.com/rsocket/rsocket-java/blob/1.0.x/rsocket-transport-netty/build.gradle)

ronenhamias (approved):

PR #481, opened 2018-12-19
Pluggable exceptions mapper

dmytro-lazebnyi:
### Please note, API of `Microservices` is changed.

Example of usage:

https://github.com/scalecube/scalecube-services/blob/6d03ac9800427ea847ddd2d274cce4a4cba14642/services-examples/src/main/java/io/scalecube/services/examples/exceptions/ExceptionMapperExample.java#L23-L66


services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -42,11 +51,13 @@ public ServiceMethodInvoker(Method method, Object service, MethodInfo methodInfo
    */
   public Mono<ServiceMessage> invokeOne(
       ServiceMessage message, BiFunction<ServiceMessage, Class<?>, ServiceMessage> dataDecoder) {
-    return Mono.defer(
+    Supplier<Mono<ServiceMessage>> invoke =
```
ronenhamias: is there more clean way to do it?

dmytro-lazebnyi: This ugly `Supplier` is added only to avoid `checkstyle` alert regarding indentation.

Unfortunately, I didn't find a good solution for this part of the code to get rid of `Supplier` variable.

The alternative solution is to move it to separate private method like below. But I don't think this is a better approach. 
What do you think?

```java
public Mono<ServiceMessage> invokeOne(
      ServiceMessage message, BiFunction<ServiceMessage, Class<?>, ServiceMessage> dataDecoder) {
    return Mono.defer(() -> invoke(message, dataDecoder))
        .onErrorResume(throwable -> Mono.just(errorMapper.toMessage(throwable)));
  }

  private Mono<ServiceMessage> invoke(
      ServiceMessage message, BiFunction<ServiceMessage, Class<?>, ServiceMessage> dataDecoder) {
    Object request = toRequest(message, dataDecoder);
    return Mono.from(invoke(request)).map(this::toResponse);
  }
```

dmytro-lazebnyi: I can move it to separate private method like:

```java
public Mono<ServiceMessage> invokeOne(
      ServiceMessage message, BiFunction<ServiceMessage, Class<?>, ServiceMessage> dataDecoder) {
    return Mono.defer(() -> invoke(message, dataDecoder))
        .onErrorResume(throwable -> Mono.just(errorMapper.toMessage(throwable)));
  }

  private Mono<ServiceMessage> invoke(
      ServiceMessage message, BiFunction<ServiceMessage, Class<?>, ServiceMessage> dataDecoder) {
    Object request = toRequest(message, dataDecoder);
    return Mono.from(invoke(request)).map(this::toResponse);
  }
```

ronenhamias: what is the checkstyle alert maybe we can remove it?

dmytro-lazebnyi: > what is the checkstyle alert

Something like this:

```
'lambda arguments' has incorrect indentation level 12, expected level should be one of the following: 6, 8. (55:0) [IndentationCheck]
'block' child has incorrect indentation level 14, expected level should be one of the following: 8, 10, 12. (56:0) [IndentationCheck]
'block' child has incorrect indentation level 14, expected level should be one of the following: 8, 10, 12. (57:0) [IndentationCheck]
'block rcurly' has incorrect indentation level 12, expected level should be one of the following: 6, 8. (58:0) [IndentationCheck]
```

segabriel (approved):
:+1:

artem-v (approved):

artem-v:
+1 after review. Thanks @dmytro-lazebnyi .

PR #487, opened 2018-12-24
Enhance API of `ServiceMessage` and `ClientMessage`

dmytro-lazebnyi:


services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
```diff
@@ -162,11 +162,33 @@ public boolean hasData(Class<?> dataClass) {
     }
   }
 
+  /**
+   * Describes whether the client message is an error.
```
artem-v: Just _message_ why exactly _client_ ?

services-gateway-clientsdk/src/main/java/io/scalecube/services/gateway/clientsdk/ClientMessage.java
```diff
@@ -72,11 +72,33 @@ public boolean hasData(Class<?> dataClass) {
     return dataClass.isPrimitive() ? hasData() : dataClass.isInstance(data);
   }
 
+  /**
+   * Describes whether the client message is an error.
```
artem-v: Just message, why exactly _clietn_ ?

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
dmytro-lazebnyi: Fixed.

services-gateway-clientsdk/src/main/java/io/scalecube/services/gateway/clientsdk/ClientMessage.java
dmytro-lazebnyi: Fixed.

PR #490, opened 2018-12-26
Fixed remote JMX support

### LINKED ISSUE #489: remote access to JMX doesn't work
jconsole can't connect to JMX.

segabriel:
Closes #489


artem-v (approved):
+1.  
Could you please do the same for `scalecube-seed` .. Just double check we dont have similar issue there.

PR #495, opened 2018-12-29
Added JMX support for microservices instance

segabriel:


ronenhamias:
i would suggest to look on zipkin tracing vs jmx

segabriel:
@ronenhamias 
> i would suggest to look on zipkin tracing vs jmx

This PR is not about tracing - we need to have some mechanism to get to know which settings were applied when microservices starts.

services/src/main/java/io/scalecube/services/MicroservicesMBean.java
```diff
@@ -0,0 +1,26 @@
...
+
+  Map<String, InetSocketAddress> getGatewayAddresses();
+
+  String getServiceEndpoint();
+
+  List<String> getRecentServiceDiscoveryEvents();
+
+  List<String> getServiceEndpoints();
+
+  List<String> getServiceReferences();
```
artem-v: Let's remove this. `getServiceEndpoints()` has it all.

services/src/main/java/io/scalecube/services/MicroservicesMBean.java
```diff
@@ -0,0 +1,26 @@
...
+
+  String getServiceTransport();
+
+  String getServiceDiscovery();
+}
```
artem-v, line 22: Let's return even single String as  List<String>   (for conveniet  rendering in jconsole)

artem-v (dismissed):

artem-v (approved):

PR #498, opened 2019-01-06
Security release

aharonha:


aharonha:
We should do that first in [scalecube/cluster](https://github.com/scalecube/scalecube-cluster/pull/156)

artem-v (approved):

PR #504, opened 2019-01-17
Feature/tests for all service transport implementations

ronenhamias:


artem-v:
I suggest to postpone this PR. But branch can be kept.

PR #511, opened 2019-01-29
Refactoring Microservices builder approach

segabriel:


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -57,15 +60,15 @@
...
- *   <li>location transparency and discovery of service instances.
- *   <li>fault tolerance using gossip and failure detection.
- *   <li>share nothing - fully distributed and decentralized architecture.
- *   <li>Provides fluent, java 8 lambda apis.
- *   <li>Embeddable and lightweight.
- *   <li>utilizes completable futures but primitives and messages can be used as well completable
- *       futures gives the advantage of composing and chaining service calls and service results.
- *   <li>low latency
- *   <li>supports routing extensible strategies when selecting service end-points
+ * <li>location transparency and discovery of service instances.
```
segabriel: Please check your code style - it seems wrong.

ronenhamias:
need to update examples and docs

services-examples/src/main/java/io/scalecube/services/examples/helloworld/Example1.java
```diff
@@ -20,13 +20,14 @@
    * @param args ignored
    */
   public static void main(String[] args) {
+    Microservices ms = new Microservices();
```
segabriel: let's don't demonstrate bad practice, use everytime new instance of Microservices.

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -104,85 +107,210 @@
...
+
+  public Microservices discovery(ServiceDiscovery discovery) {
+    Microservices msNew = new Microservices(this);
+    msNew.discovery = discovery;
+    return msNew;
+  }
+
+  public Microservices discovery(Consumer<ServiceDiscoveryConfig.Builder> discoveryOptions) {
+    Microservices msNew = new Microservices(this);
+    msNew.discoveryOptions = discoveryOptions;
```
segabriel, line 281: I think it looks wrong a bit. I'd like to suppose the following solution for it:
```java
    if (msNew.discoveryOptions != null) {
      msNew.discoveryOptions = msNew.discoveryOptions.andThen(discoveryOptions);
    } else {
      msNew.discoveryOptions = discoveryOptions;
    }
```
It will give us  an opportunity do the next one:
```java
        Microservices.builder()
            .discovery(options -> options.seeds(seed.discovery().address()))
            .discovery(options -> options.port(123))
            .discovery(options -> options.memberHost("member"))
```

ronenhamias: i dont like this line:

```java
if (msNew.discoveryOptions != null) {
```

in what case it can be null?

segabriel: @ronenhamias do you mind of null value or compose two consumers?
so now, the null value of this consumer is legal state. For example, when we create a seed it doesn't have discovery option consumer at all.

ronenhamias: i think that in either case it should not be null.

segabriel:
let's implement https://github.com/scalecube/scalecube-services/issues/512 and change all null with NotNullObjects like: `NoopDiscoveryOptions`

services/src/main/java/io/scalecube/services/Microservices.java
dmytro-lazebnyi: @segabriel @ronenhamias 
Things what you are discussing here is not a part of this pull request.

This pull request is about _refactoring_ which generally means _change code structure without changing behavior_.
@segabriel your proposition https://github.com/scalecube/scalecube-services/pull/511#discussion_r252613337 should be discussed out of this pull request as a separate issue, because in the previous implementation (with Builder) we had totally the same behavior regarding null-checking.

artem-v:
Adjust checkstyle plugin in your IDEA and google formatter:  https://github.com/scalecube/scalecube-parent/blob/develop/DEVELOPMENT.md#set-up-intellij-idea

artem-v:
Why `doShutdown()` is being exposed as public method. 
Method `shutdown()` is left exposed as well.  Why? 
Only 1 must be accessible for the client.

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -104,85 +107,313 @@
...
-    this.metrics = builder.metrics;
-    this.tags = new HashMap<>(builder.tags);
-    this.serviceProviders = new ArrayList<>(builder.serviceProviders);
-    this.serviceRegistry = builder.serviceRegistry;
-    this.methodRegistry = builder.methodRegistry;
-    this.gatewayBootstrap = builder.gatewayBootstrap;
-    this.discovery = builder.discovery;
-    this.discoveryOptions = builder.discoveryOptions;
-    this.transportBootstrap =
+  private static final Logger log = LoggerFactory.getLogger(Microservices.class);
```
artem-v: Call it `LOGGER`

artem-v (changes requested):

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -104,85 +107,313 @@
...
+
+  /**
+   * Start deferred {@code Microservices} with {@code doStart()} method.
+   *
+   * @return
+   */
+  public Mono<Microservices> doStart() {
+    return Mono.defer(
+        () ->
+            this.transportBootstrap
```
segabriel: let's create here a new instance of Microservices and right here subscribe to shutdown hook. So it will give us that each call of this method will return a new instance of Microservices and _only_ in the success case it will subscribe to shutdown hooks. And we shouldn't support `doStart` and `start` mono processors at all. So let's remove them.

@ronenhamias @dmytro-lazebnyi @stingion what do you think about it?

ronenhamias: +1

stingion:
> Why `doShutdown()` is being exposed as public method.
> Method `shutdown()` is left exposed as well. Why?
> Only 1 must be accessible for the client.

Made `shutdown()` to be "private"

ronenhamias (changes requested):
1. is that how it suppose to be?
```java
public Microservices metrics(Metrics metrics) {
    Microservices msNew = new Microservices(this);
    msNew.metrics = metrics;
    return msNew;
  }
```

2. update all docs and user references across the globe including  the javadocs and website 

explain this change to the user how exactlly they should use the lib.
```java
 *  // Build a microservices cluster instance:
 *  Microservices microservices = Microservices.builder()
 *       // Introduce GreetingServiceImpl pojo as a micro-service:
 *      .services(new GreetingServiceImpl())
 *      .startAwait();
 *
 *  // Create microservice proxy to GreetingService.class interface:
 *  GreetingService service = microservices.call().create()
 *      .api(GreetingService.class);
 *
```
fix readme and user guide.
https://github.com/scalecube/scalecube-services/blob/develop/LICENSE.txt

3. what happen if someone comes and write the following line:

```
Microservices ms = new Microservices();
ms.metrics(metrics);
ms.start()
```

he expected that his ms runs with metrics but actually its not :)

stingion:
> 1. why we need newInstance(...) ? - please remove.
> 2. update all docs and user references across the globe including  the javadocs and website
> 
> explain this change to the user how exactlly they should use the lib.
> 
> ```java
>  *  // Build a microservices cluster instance:
>  *  Microservices microservices = Microservices.builder()
>  *       // Introduce GreetingServiceImpl pojo as a micro-service:
>  *      .services(new GreetingServiceImpl())
>  *      .startAwait();
>  *
>  *  // Create microservice proxy to GreetingService.class interface:
>  *  GreetingService service = microservices.call().create()
>  *      .api(GreetingService.class);
>  *
> ```
> 
> fix readme and user guide.
> https://github.com/scalecube/scalecube-services/blob/develop/LICENSE.txt

1. Remove redundant instance creation methods (factories)

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -225,19 +432,19 @@ public InetSocketAddress gatewayAddress(String name, Class<? extends Gateway> ga
...
+    shutdown.onComplete();
   }
 
   /**
    * Shutdown instance and clear resources.
    *
    * @return result of shutdown
    */
-  public Mono<Void> shutdown() {
+  public Mono<Void> doShutdown() {
```
segabriel, line 444: please rename it to `shutdown`

services/src/test/java/io/scalecube/services/StreamingServiceTest.java
```diff
@@ -24,11 +24,12 @@
   /** Setup. */
   @BeforeAll
   public static void setup() {
-    gateway = Microservices.builder().startAwait();
+    Microservices ms = new Microservices();
```
segabriel, line 27: please inline it (just remove the local variable)

services/src/test/java/io/scalecube/services/ServiceTransportTest.java
```diff
@@ -32,11 +31,12 @@
   /** Setup. */
   @BeforeEach
   public void setUp() {
-    gateway = Microservices.builder().startAwait();
+    Microservices ms = new Microservices();
```
segabriel, line 34: please inline it (just remove the local variable)

services/src/test/java/io/scalecube/services/ServiceRegistryEventsTest.java
```diff
@@ -18,30 +17,26 @@ public void test_added_removed_registration_events() {
 
     List<ServiceDiscoveryEvent> events = new ArrayList<>();
 
-    Microservices seed = Microservices.builder().startAwait();
+    Microservices ms = new Microservices();
```
segabriel, line 20: please inline it (just remove the local variable)

services/src/test/java/io/scalecube/services/ServiceRegistryEventsTest.java
```diff
@@ -18,30 +17,26 @@ public void test_added_removed_registration_events() {
...
-    assertEquals(4, events.size());
-    assertEquals(Type.REGISTERED, events.get(0).type());
-    assertEquals(Type.REGISTERED, events.get(1).type());
-    assertEquals(Type.UNREGISTERED, events.get(2).type());
-    assertEquals(Type.UNREGISTERED, events.get(3).type());
+    seed.doShutdown().block(Duration.ofSeconds(6));
+  }
 
-    seed.shutdown().block(Duration.ofSeconds(6));
+  private Microservices getMs(Microservices ms, Microservices seed) {
```
segabriel, line 37: remove it, it looks so sophisticated.

services/src/test/java/io/scalecube/services/routings/RoutersTest.java
```diff
@@ -187,7 +186,7 @@ public void test_tag_request_selection_logic() {
           service.requestOne(GREETING_REQUEST_REQ2, GreetingResponse.class).block(timeout).data();
       GreetingResponse resultForJoe =
           service.requestOne(GREETING_REQUEST_REQ, GreetingResponse.class).block(timeout).data();
-      assertEquals("1", resultForJoe.sender());
+      assertEquals("2", resultForJoe.sender());
```
ronenhamias, line 189: i am almost sure we should not change the assertions as result of this change

ronenhamias (changes requested):

ronenhamias:
it is unfortunate but i guess we have to close this pull request  because its creating not a clean and clear api and not much of a clear value about this change.

PR #515, opened 2019-02-13
Bug fix with http method without param in HttpGatewayAcceptor

### LINKED ISSUE #513: Method not being called when making HTTP request to @ServiceMethod that have no input params
Method not being called when making HTTP request to @ServiceMethod that have no input params, response is empty and no error in log.

Example:

**Definition**
`@ServiceMethod("getVersion")
    Mono<String> getVersion();`

**Implementation**
`@Override
    public Mono<String> getVersion() {
        return Mono.just("version 1");
    }`

**Request:**

- POST

- URL: localhost:8080/adminservice/getVersion

- Headers: Content-Type:application/json

- Body: <empty>

stingion:
Fix #513

### DIFF
--- a/services-gateway-http/src/main/java/io/scalecube/services/gateway/http/HttpGatewayAcceptor.java
+++ b/services-gateway-http/src/main/java/io/scalecube/services/gateway/http/HttpGatewayAcceptor.java
@@ -64,8 +64,7 @@ public class HttpGatewayAcceptor
     return httpRequest
         .receive()
         .aggregate()
-        .switchIfEmpty(
-            Mono.defer(() -> ByteBufMono.just(ByteBufAllocator.DEFAULT.compositeBuffer())))
+        .switchIfEmpty(Mono.defer(() -> ByteBufMono.just(Unpooled.EMPTY_BUFFER)))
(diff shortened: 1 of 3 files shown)


services-gateway-http/src/main/java/io/scalecube/services/gateway/http/HttpGatewayAcceptor.java
```diff
@@ -64,6 +64,8 @@
     return httpRequest
         .receive()
         .aggregate()
+        .switchIfEmpty(
+            Mono.defer(() -> ByteBufMono.just(ByteBufAllocator.DEFAULT.compositeBuffer())))
```
segabriel: what is composite buffer maybe we need an empty buffer?

ronenhamias (approved):
+1

services-gateway-http/src/main/java/io/scalecube/services/gateway/http/HttpGatewayAcceptor.java
stingion: Changed

segabriel (approved):
:+1:

PR #516, opened 2019-02-22
Fix broken links in README.md

IgorPerikov:
I wasn't sure about the exact way of how to replace the broken links, so I fixed them in a way, so they refer to some pages that can be found and provide some useful information (from my perspective of view at least)


ronenhamias (approved):

ronenhamias:
thank you

PR #518, opened 2019-03-04
Refactored startup for ServiceDiscovery

artem-v:


services-discovery/src/main/java/io/scalecube/services/discovery/ClusterAddresses.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Converts one address to another.
+   *
+   * @param address scalecube services address
+   * @return scalecube cluster address
+   */
+  public static io.scalecube.transport.Address toAddress(
+      io.scalecube.services.transport.api.Address address) {
+    return io.scalecube.transport.Address.create(address.host(), address.port());
```
ronenhamias, line 20: we better consolidate all address to seperate jar so it can be used accross the modules

services-examples/src/main/java/io/scalecube/services/examples/exceptions/ExceptionMapperExample.java
```diff
@@ -15,6 +21,9 @@
   public static void main(String[] args) throws InterruptedException {
     Microservices ms1 =
         Microservices.builder()
+            .discovery(
+                (serviceRegistry, serviceEndpoint) ->
+                    new ScalecubeServiceDiscovery(serviceRegistry, serviceEndpoint).start())
```
ronenhamias: IMHO i think we can take the time and not pass the service registry to the service discovery
we should pass services we want discovery to publish when starting 
the bridge between service registry and discovery should be in microservices.

artem-v: This is refactoring stage, I keep functionalty without changes , without introducing new concepts.

services-discovery/src/main/java/io/scalecube/services/discovery/ClusterAddresses.java
artem-v: This ClusterAddress utility was created  for purpose  of  integration  Microservices  type  Address and  scalecube  Cluster  Address, nothing more

services-benchmarks/src/main/java/io/scalecube/services/benchmarks/transport/BenchmarkServiceState.java
```diff
@@ -27,12 +30,24 @@ public BenchmarkServiceState(BenchmarkSettings settings, Object... services) {
 
   @Override
   public void beforeAll() {
-    seed = Microservices.builder().metrics(registry()).startAwait();
+    seed =
+        Microservices.builder()
+            .metrics(registry())
+            .discovery(
+                (serviceRegistry, serviceEndpoint) ->
```
segabriel: it looks so convoluted, do we have another way to specify necessary options?

services-examples/src/main/java/io/scalecube/services/examples/BootstrapExample.java
```diff
@@ -27,6 +33,9 @@ public static void main(String[] args) throws Exception {
     System.out.println("Start gateway");
     Microservices gateway =
         Microservices.builder()
+            .discovery(
+                (serviceRegistry, serviceEndpoint) ->
+                    new ScalecubeServiceDiscovery(serviceRegistry, serviceEndpoint).start())
```
segabriel: here we invoke `start` and in other places of this class we don't do it, can we get rid of this invoking?

segabriel: let's apply your idea with  a supplier -> when microservices start it will invoke `start` itself

services-discovery/src/main/java/io/scalecube/services/discovery/ClusterAddresses.java
ronenhamias: its created already several times and feels more like copy paste then solution

io.scalecube.transport.Address
io.scalecube.services.transport.api.Address

services-benchmarks/src/main/java/io/scalecube/services/benchmarks/transport/BenchmarkServiceState.java
ronenhamias: i agree maybe its not new functionalty but it offers new convoluted feature

artem-v: @segabriel 
> it looks so convoluted, do we have another way to specify necessary options?

i will address only constructive criticism

services-discovery/src/main/java/io/scalecube/services/discovery/ClusterAddresses.java
artem-v: Point of this refactoring is not to deal with messy Address types. Im very well aware of two Address classes. In this PR focus was on something else.

services-examples/src/main/java/io/scalecube/services/examples/BootstrapExample.java
artem-v: > here we invoke start and in other places of this class we don't do it, can we get rid of this invoking?

I will move `start()`  to `ServiceDiscovery`

segabriel (approved):

segabriel:
:+1:

artem-v:
Thanks for review @segabriel , @ronenhamias

Issue #519, opened 2019-03-05
Decouple Service Registry from discovery


ronenhamias:
```
 node =
        Microservices.builder()	        Microservices.builder()
            .metrics(registry())	            .metrics(registry())
            .discovery(options -> options.seeds(seed.discovery().address()))	            .discovery(
                (serviceRegistry, serviceEndpoint) ->
                    new ScalecubeServiceDiscovery(serviceRegistry, serviceEndpoint)
                        .options(opts -> opts.seedMembers(seedAddress)))
            .services(services)	            .services(services)
            .startAwait();	            .startAwait();
```
we should not pass service registry to service discovery.
```
(serviceRegistry, serviceEndpoint) ->
```
the bridge should be done inside microservices.

PR #521, opened 2019-03-06
Issue519 decouple service registry from discovery

artem-v:
Fixes https://github.com/scalecube/scalecube-services/issues/519


services-discovery/src/main/java/io/scalecube/services/discovery/ScalecubeServiceDiscovery.java
```diff
@@ -27,41 +26,39 @@
...
+  private static final int EVENT_BUFFER_SIZE = 65536;
+
   private final ServiceEndpoint endpoint;
   private final ClusterConfig clusterConfig;
 
   private Cluster cluster;
 
-  private final DirectProcessor<ServiceDiscoveryEvent> subject = DirectProcessor.create();
+  private final EmitterProcessor<ServiceDiscoveryEvent> subject =
+      EmitterProcessor.create(EVENT_BUFFER_SIZE);
```
segabriel: set `autoCancel` to false here

segabriel (approved):
just set `autoCancel` to false in `emmitterprocessor`

ronenhamias (approved):

services-discovery/src/main/java/io/scalecube/services/discovery/ScalecubeServiceDiscovery.java
artem-v: Done

PR #522, opened 2019-03-07
Refactor gateway bootstrap process

artem-v:


services-examples/src/main/java/io/scalecube/services/examples/BootstrapExample.java
```diff
@@ -34,21 +33,10 @@ public static void main(String[] args) throws Exception {
...
-                    .port(8181)
-                    .build()) // override default port
-            .gateway(
-                GatewayConfig.builder("ws", WebsocketGatewayStub.class)
-                    .port(9191)
-                    .addOption(
-                        WebsocketGatewayStub.WS_SPECIFIC_OPTION_NAME,
-                        "500") // override default value of specific option
-                    .build())
+            .gateway(opts -> new HttpGatewayStub(opts.id("http").port(8181)))
```
ronenhamias: lets find better name then stub

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -620,13 +674,6 @@ private JmxMBeanBootstrap(Microservices microservices) {
       return Collections.singletonList(microservices.discovery().address().toString());
     }
 
-    @Override
-    public Collection<String> getGatewayAddresses() {
```
segabriel: Please add gateway binding info in JMX bean.

segabriel (approved):
looks better

services-examples/src/main/java/io/scalecube/services/examples/BootstrapExample.java
artem-v: Changed name to end with postfix -- _Example_ instead of _Stub_

services/src/main/java/io/scalecube/services/Microservices.java
artem-v: Done

PR #541, opened 2019-03-27
Fix to Block Microservices on termination signal #529

### LINKED ISSUE #529: Block Microservices on termination signal
After microservices has started block it on termination signal (such as SIGTERM for example).
Upon receiving exiting signal do microservices shutdown.

http://man7.org/linux/man-pages/man7/signal.7.html
https://docs.oracle.com/javase/10/troubleshoot/handle-signals-and-exceptions.htm#JSTGD345

artem-v:
Fixed #529

### DIFF
--- a/services/src/main/java/io/scalecube/services/Microservices.java
+++ b/services/src/main/java/io/scalecube/services/Microservices.java
@@ -488,7 +488,7 @@ public class Microservices {
                   .doFinally(
                       s -> {
                         if (!gateways.isEmpty()) {
-                          LOGGER.info("Gateways has been stopped");
+                          LOGGER.info("Gateways have been stopped");
                         }
                       }));
     }
--- a/services/src/main/java/io/scalecube/services/Microservices.java
+++ b/services/src/main/java/io/scalecube/services/Microservices.java
@@ -40,9 +40,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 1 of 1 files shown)


services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -448,7 +488,7 @@ private GatewayBootstrap addFactory(Function<GatewayOptions, Gateway> factory) {
                   .doFinally(
                       s -> {
                         if (!gateways.isEmpty()) {
-                          LOGGER.info("Gateways have been stopped");
+                          LOGGER.info("Gateways has been stopped");
```
dmytro-lazebnyi: `have`

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -244,11 +258,37 @@ public ServiceDiscovery discovery() {
...
   public Mono<Void> shutdown() {
     return Mono.defer(
-        () ->
-            Mono.whenDelayError(
-                discoveryBootstrap.shutdown(),
-                gatewayBootstrap.shutdown(),
-                transportBootstrap.shutdown()));
+        () -> {
+          shutdown.onComplete();
+          return onShutdown;
```
segabriel, line 263: if we have the separate method, why do we need to return anything? I propose to return just `void` from this method.

artem-v: Because I don't want to change clients all around the code.

segabriel (approved):
great job!

dmytro-lazebnyi (approved):

PR #554, opened 2019-05-10
update rsocket version 0.12.2-RC2

ronenhamias:


segabriel (approved):
Great! I found that exactly this version of the rsocket the spring uses
https://github.com/spring-projects/spring-framework/blob/06c3a058f4b5be96e7547d5d9bd36c49a4528b2b/spring-messaging/spring-messaging.gradle#L10

But it's still release candidate

PR #574, opened 2019-05-23
Add groups to service discovery

artem-v:


segabriel (approved):
But I don't understand it fully

ronenhamias (approved):

PR #598, opened 2019-06-10
Registry print warning message, if service have overridden

eutkin:
If we have two implementation of single interface and the one is override other, logger print warn message

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
+++ b/services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
@@ -4,17 +4,15 @@ import io.scalecube.services.api.ServiceMessage;
 import io.scalecube.services.exceptions.BadRequestException;
 import io.scalecube.services.exceptions.ServiceProviderErrorMapper;
 import io.scalecube.services.transport.api.ServiceMessageDataDecoder;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Stream;
-
-import org.reactivestreams.Publisher;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
 
 import static java.util.stream.Collectors.joining;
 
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/scalecube/services/methods/ServiceMethodRegistryImpl.java
```diff
@@ -36,10 +42,14 @@ public void registerService(
...
-                              methodInfo.qualifier(),
-                              new ServiceMethodInvoker(
-                                  method, serviceInstance, methodInfo, errorMapper, dataDecoder));
+                          String qualifier = methodInfo.qualifier();
+                          ServiceMethodInvoker invoker = new ServiceMethodInvoker(
+                                  method, serviceInstance, methodInfo, errorMapper, dataDecoder);
+
+                          if (methodInvokers.containsKey(qualifier)) {
+                            LOGGER.warn("<{}> for API <{}> will be replaced by the following methodInvoker: <{}>", methodInvokers.get(qualifier), qualifier, invoker);
+                          }
```
artem-v, line 45: Throw  exception (`IllegalStateException` should be fine) instead of allowing client to progress with wrong state.

segabriel:
Please apply google style. See https://github.com/scalecube/scalecube-parent/blob/develop/DEVELOPMENT.md#code-style

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -4,126 +4,143 @@
...
+    }
+
+    @Override
+    public String toString() {
+        String prefix = service.getClass().getCanonicalName() + "#" + method.getName() + "(";
+        return Stream
+                .of(method.getParameters())
+                .map(Parameter::getType)
+                .map(Class::getSimpleName)
+                .collect(joining(", ", prefix, ")"));
```
artem-v: This .toString is confusing. Make it simpler. At line 144 use just `"("` string.  After that concat `service.getClass().getCanonicalName() + "#" + method.getName()` with that updated joined string.

services/src/main/java/io/scalecube/services/methods/ServiceMethodRegistryImpl.java
```diff
@@ -36,10 +42,14 @@ public void registerService(
...
-                          methodInvokers.put(
-                              methodInfo.qualifier(),
-                              new ServiceMethodInvoker(
-                                  method, serviceInstance, methodInfo, errorMapper, dataDecoder));
+                          String qualifier = methodInfo.qualifier();
+                          ServiceMethodInvoker invoker = new ServiceMethodInvoker(
+                                  method, serviceInstance, methodInfo, errorMapper, dataDecoder);
+
+                          if (methodInvokers.containsKey(qualifier)) {
+                            LOGGER.warn("<{}> for API <{}> will be replaced by the following methodInvoker: <{}>", methodInvokers.get(qualifier), qualifier, invoker);
```
artem-v: This has to be `LOGGER.error()`  ;   use this string --
 `Existing methodInvoker ({}) for api '{}' is about to be replaced by new methodInvoker ({})`

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
eutkin: I don't understand, sorry. 
```java
service.getClass().getCanonicalName() + "#" + method.getName() + Stream
                .of(method.getParameters())
                .map(Parameter::getType)
                .map(Class::getSimpleName)
                .collect(joining(", ", "(", ")"));
```

like this? 
 P.S. извини, я просто не владею разговорным англ языком

artem-v: > like this?

Да, норм

artem-v:
Fix check style   -- https://travis-ci.org/scalecube/scalecube-services/builds/543689203?utm_source=github_status&utm_medium=notification

segabriel (approved):

segabriel:
@eutkin Thanks for your contribution!

PR #600, opened 2019-06-11
Added tags definitions in @Service, @ServiceMethod annotations

eutkin:


eutkin:
Example:

```java
@Service(tags = {"tagA", "a", "tagB", "b"})
@FunctionalInterface
public interface TagService {

  @ServiceMethod(tags = {"methodTagA", "a"})
  Flux<String> upperCase(Flux<String> input);

}
```

Way to set tags can be discussed

ronenhamias:
@eutkin  i was considering this feature back in the days and was trying to think about use-cases when its valuable since usually when setting tags in annotations they are become static (hard coded)
and when setting tags at bootstrap phase they are driven by configuration. 

in general with scalecube we separate the bootstrap of microservices from the business logic this helps to keep a clean code and flexibility.

thinking about it together with you... how about an approach can be maybe to introduce a new Annotation ```@Tags``` this way we introduce this new annotation once and not need to change ```@Service``` and ```@ServiceMethod```

```java
@Service()
@Tags("tag-a","a", "tag-b","b" )
public interface GreetingService {

  @ServiceMethod()
  @Tags("m-tag-a","a", "m-tag-b","b" )
  Flux<String> hello(Flux<String> input);

}
```

question the comes to mind is what if user set tags in bootstrap and in @Tags annotation can they conflict? what overrides who?

services/src/main/java/io/scalecube/services/ServiceScanner.java
```diff
@@ -31,7 +42,10 @@ private ServiceScanner() {
               List<ServiceMethodDefinition> actions =
                   Arrays.stream(serviceInterface.getMethods())
                       .filter(method -> method.isAnnotationPresent(ServiceMethod.class))
-                      .map(method -> new ServiceMethodDefinition(Reflect.methodName(method)))
+                      .map(
+                          method ->
+                              new ServiceMethodDefinition(Reflect.methodName(method))
+                                  .setTags(Reflect.serviceMethodTags(method)))
```
artem-v: Remove  method `setTags` (as well as `setAction`  method) and use another constructor `ServiceMethodDefinition(String action, Map<String, String> tags)`

services/src/test/java/io/scalecube/services/routings/RoutersTest.java
```diff
@@ -2,33 +2,27 @@
...
 import io.scalecube.services.discovery.ScalecubeServiceDiscovery;
 import io.scalecube.services.discovery.api.ServiceDiscovery;
 import io.scalecube.services.routing.RandomServiceRouter;
 import io.scalecube.services.routing.Routers;
-import io.scalecube.services.routings.sut.CanaryService;
-import io.scalecube.services.routings.sut.DummyRouter;
-import io.scalecube.services.routings.sut.GreetingServiceImplA;
-import io.scalecube.services.routings.sut.GreetingServiceImplB;
-import io.scalecube.services.routings.sut.WeightedRandomRouter;
+import io.scalecube.services.routings.sut.*;
```
artem-v: Adjust your IDE not to put asterisk  in imports.

services/src/main/java/io/scalecube/services/ServiceScanner.java
```diff
@@ -21,8 +24,16 @@ private ServiceScanner() {
    */
   public static List<ServiceRegistration> scanServiceInfo(ServiceInfo serviceInfo) {
     return Arrays.stream(serviceInfo.serviceInstance().getClass().getInterfaces())
-        .map(serviceInterface -> new InterfaceInfo(serviceInterface, serviceInfo.tags()))
-        .filter(interfaceInfo -> interfaceInfo.serviceInterface.isAnnotationPresent(Service.class))
+        .filter(serviceInterface -> serviceInterface.isAnnotationPresent(Service.class))
+        .map(
+            serviceInterface -> {
+              Map<String, String> userTags = serviceInfo.tags();
```
artem-v: In contrast to `apiTags` where it's clear what is it, the `userTags` var name sounds confusing. Lets rename it to `clientTags` or `serviceTags` or `programTags`. Because we have to designate that this is _programmatic_ tags coming from ServiceInfo form the client.

services/src/main/java/io/scalecube/services/ServiceScanner.java
```diff
@@ -21,8 +24,16 @@ private ServiceScanner() {
...
+              Map<String, String> apiTags = Reflect.serviceTags(serviceInterface);
+              Map<String, String> buffer = new HashMap<>(apiTags);
+              // user tags override tags from @Service
+              buffer.putAll(userTags);
```
artem-v: Looks like you doing this correctly, u override annotation tags with programmatic tags. Also look at this `io.scalecube.services.ServiceReference#mergeTags`  for details. This is FYI.

artem-v (changes requested):

artem-v:
@snripa 
@dmytro-lazebnyi 
Guys take a look at this comment -- https://github.com/scalecube/scalecube-services/pull/600#issuecomment-501146445.  `tags` in attribute of existing annotation vs  seprate `@Tags` annottation. WDYT?

segabriel:
@artem-v better reuse existing annotations.

dmytro-lazebnyi:
@artem-v 
> tags in attribute of existing annotation vs seprate @Tags annottation. WDYT?

I think the `tags` attribute of existing annotations is better than the new `@Tags` annotation.

Tags are part of service (service method) definition and so it should be used along with `@Service` or `@ServiceMethod`. 
Thus, having separate `@Tags` annotation probably could add a false expectation that this annotation can be used without `@Service` or `@ServiceMethod`.

ronenhamias:
@artem-v @eutkin @segabriel  @dmytro-lazebnyi 

thinking about it again... yhea maybe its better to reuse existing annotations.

services/src/main/java/io/scalecube/services/ServiceScanner.java
eutkin: i have seen

eutkin:
I see no reason to create a lot of annotations to describe the service. The user is not the fact that he finds out about the existence of @Tags.

artem-v (approved):

artem-v:
Merged, thanks @eutkin

PR #601, opened 2019-06-14
DI for RSocketTransport

eutkin:
I added DI support for RSocketTransport. This was necessary to support DI in the project as a whole, as an alternative to the SPI mechanism.

RSocketServiceTransport depends on HeadersCodec. I want to use JacksonCodec. JaсsonCodec depends on ObjectMapper. The user should be able to configure the ObjectMapper.SPI does not allow this.


artem-v (changes requested):
https://github.com/scalecube/scalecube-parent/blob/develop/DEVELOPMENT.md

Adjust your formatter

services-transport-parent/services-transport-rsocket/src/main/java/io/scalecube/services/transport/rsocket/RSocketServiceTransport.java
```diff
@@ -14,29 +14,65 @@
  */
 public class RSocketServiceTransport implements ServiceTransport {
 
+  /**
+   * Default Instance.
+   *
```
artem-v: Format this file.

segabriel (approved):

artem-v (approved):

PR #602, opened 2019-06-20
TypeSafe Service Transport Builder.

eutkin:
Motivation:

Very often, TransportResources and Client / Server Transport are connected. A specific Client / Server Transport implementation needs a specific implementation. The current API did not control type safety.

Example:

```java
TransportResources nettyTcpResources = new RSocketTransportResources();
Microservices.builder()
  .transport(opts -> opts.resources(nettyTcpResources)
                         .client(RSocketServiceTransport::client)
                         .server(RSocketServiceTransport::server)
 );

// Next:
TransportResources nettyWebsocketResources = new WebsocketResources();

// throw ClassCastException: WebsocketResources is not RSocketTransportResources
Microservices.builder()
  .transport(opts -> opts.resources(nettyWebsocketResources )
                         .client(RSocketServiceTransport::client)
                         .server(RSocketServiceTransport::server)
 );

```

Default value of transportResource is null (see `Supplier<TransportResources> resource = () -> null`. If you don't call Builder#transport, you will get NPE.

Current API is deprecated.

New API:

```
RSocketServiceTransport -> RSocketServiceTransportFactory
```
and 

```java
Microservices seed =
        Microservices.builder()
            .setupTransport(RSocketTransportResources::new)
            .host("0.0.0.0")
            .port(8080)
            .transportFactory(RSocketServiceTransportFactory::new)
            .startAwait();
```


README.md
```diff
@@ -70,7 +70,8 @@ The example provisions 2 cluster nodes and making a remote interaction.
                 serviceEndpoint ->
                     new ScalecubeServiceDiscovery(serviceEndpoint)
                         .options(opts -> opts.seedMembers(toAddress(seed.discovery().address()))))
-            .transport(ServiceTransports::rsocketServiceTransport)
+            .setupTransport(RSocketTransportResources::new)
```
ronenhamias, line 73: why to split these methods and not just already instance of ServiceTransportBootstrap with specific resources.

eutkin: no, else pull request will be have breaking changes.

eutkin:
Disclaimer: my thoughts below do not carry the purpose of offending or offending someone.

API. A good API in my opinion should allow you to do the right things and not let them do the wrong things.
For example, if the API allows you to use a resource for RSocketTransport that does not suit it, then this is bad. You did what the API allowed you and got a ClassCastException.

``` java
TransportResources nettyTcpResources = new RSocketTransportResources ();
Microservices.builder ()
  .transport (opts -> opts.resources (nettyTcpResources)
                         .client (RSocketServiceTransport::client)
                         .server (RSocketServiceTransport::server)
 );

// Next:
TransportResources nettyWebsocketResources = new WebsocketResources ();

// throw ClassCastException: WebsocketResources is not RSocketTransportResources
Microservices.builder ()
  .transport (opts -> opts.resources(nettyWebsocketResources)
                         .client(RSocketServiceTransport::client)
                         .server(RSocketServiceTransport::server)
 );

// Throw NPE
Microservices.builder()
  .transport (opts -> opts.client (RSocketServiceTransport::client)
                            .server(RSocketServiceTransport::server)
)
```
We will consider only the API for customization of transport, `Builder#transport`.

What does the current API look like?

``` java
Microservices
            .builder()
            .transport(opt -> opt.resources (RSocketTransportResources::new)
                    .client(RSocketServiceTransport.INSTANCE::clientTransport)
                    .server(RSocketServiceTransport.INSTANCE::serverTransport)
            );
``` 
Consider the method for creating ClientTransport:

`serviceTransportBootsrap.client(Function<TransportResources, ClientTransport> factory)`

First, it may come as a surprise to the user that in `resource -> new SomeClientTransport (resources)` resource maybe is nullable (but maybe it's just me stupid).

Secondly, we expect to receive not ClientTransport, but ClientTransportFactory, which creates ClientTransport based on TransportResources. Which may not be. And which can return not quite that thread pool that you want (hello, EventLoopGroup). That is, the API tells the user that you must create your transport based on TransportResources. If you pass `resourceIgnore -> clientTransportInstance`, then it looks more like a dirty trick.

Therefore, I tried to make the API more unambiguous. The current API seems to say, hey, TransportResources is something very important, and not just an option, don't forget to put it down, guy.

About Renaming setupTransport -> transport. This will create an overload of the old method and it will break backwards compatibility (don't you think that I really wanted to call the setupTransport method?)

I think your API is suitable for people who know exactly what they are doing. And random people who do not know the details of implementation, can shoot themselves in the foot and not notice

And I will add about TransportResources. If we use DI to configure transport in scalecube, then we can’t say for sure which implementation of TransportResources will be used. Otherwise, the weak coupling principle is violated. Therefore, it is necessary either to encapsulate everything about transport in one interface / implementation, or to narrow down the range of potential implementations of TransportResources.

eutkin:
I think your API is suitable for people who know exactly what they are doing. And random people who do not know the details of implementation, can shoot themselves in the foot and not notice

eutkin:
And I will add about TransportResources. If we use DI to configure transport in scalecube, then we can’t say for sure which implementation of TransportResources will be used. Otherwise, the weak coupling principle is violated. Therefore, it is necessary either to encapsulate everything about transport in one interface / implementation, or to narrow down the range of potential implementations of TransportResources.

artem-v (changes requested):
Let's close this in favor of https://github.com/scalecube/scalecube-services/pull/607

ronenhamias:
@eutkin we can close this pull request?
i think its done in another....

eutkin:
@ronenhamias ok. No matter who solved the problem and how, it is important that it was solved

PR #613, opened 2019-06-27
Revert stable rsocket version

segabriel:


segabriel:
It seems a reason is some blocking code.
No relevant now.

artem-v (approved):

PR #615, opened 2019-06-28
Keep-alive rsocket problem and Edge cases

segabriel:
Problems:

1) We have no way to change keep-alive settings, they are adjusting in 

```java
RSocketFactory.connect()
            .keepAlive(...)
            .frameDecoder(PayloadDecoder.ZERO_COPY)
```

2) Everyone should be more careful when implementing a service interface. Need to choose the right operator (`publishOn`/`subscribeOn`) and set it in the right place to release connection/rsocket thread. So maybe it will be better to provide some mechanism to relieve it and make it automatically. Sure it should be by demand.


artem-v:
> We have no way to change keep-alive settings, they are adjusting in

Yes, for sure. Looks like we have to provide more flexible way of how to create rsocket-client-transport and rsocket-server-transport. We lil bit locked ourselves with the constructors.

artem-v:
> So maybe it will be better to provide some mechanism to relieve it and make it automatically. Sure it should be by demand.

Imho
Better to invest a time in determining conditions under which this keep-alive-issue may occur. And put (in examples module) a service which shows how to prevent keep-alive-issue.

artem-v (changes requested):
Let's add a test case where we're doing two serial service calls. For a first service call conditions for keep-alive-issue must be met, and we should get an error;  for a second call there should be no staged keep-alive .. just normal service call, and it must pass.  The point is to ensure that keep-alive-issue received for call1 is not an obstacle for normal service call2, i.e. that we can still communicate with service even though in the past there was keep-alive error.

artem-v (approved):

artem-v (changes requested):
Refactor to examples.

artem-v (approved):
Fix checkstyle and merge plz

PR #616, opened 2019-07-01
Rewriting Transport API. Continued my previous PR

eutkin:
# New Transport API for scalecube

## Features

* Separated RSocket Transport and Tcp Netty Transport

* Easy configuration

### RSocketTransport:

Creation of transport is taken out of RSocketTransport:

* ClientTransport => ClientTransportFactory

* ServerTransport => ServerTransportFactory

### Scalecube Transport:
    
ServiceTransportProvider and builders mechanism added to create and configure specific 
implementations, example `RSocketByNettyTcp`.

## API Examples

```java
ServiceTransportProvider tcpRSocketProvider = RSocketNettyTcp
        .builder()
        .customizeClient(TcpClient::noProxy)
        .customizeClient(TcpClient::secure)
        .customizeServer(tcpServer -> tcpServer.wiretap(true))
        .customizeServer(TcpServer::noSSL)
        .dataCodecs(new JacksonCodec(), new ProtostuffCodec())
        // or
        .dataCodecs(asList(new JacksonCodec(), new ProtostuffCodec()))
        .headersCodec(new JacksonCodec())
        .build();

Microservices seed =
        Microservices.builder()
            .discovery(ScalecubeServiceDiscovery::new)
            .transport(opt -> opt
                .transportProvider(tcpRSocketProvider)
                .host("host")
                .port(9000)
            )
            .startAwait();
```

PR #619, opened 2019-07-01
Fine grained service transport construction

### LINKED ISSUE #608: ServerTransport.bind() must be decoupled from host/port in signature
In function `io.scalecube.services.transport.api.ServerTransport#bind`  there must be no host/port in signature.  Plus `bind()` must be renamed to  `start()`.  Remained dependency on `io.scalecube.services.methods.ServiceMethodRegistry` is ok.

artem-v:
Fixes https://github.com/scalecube/scalecube-services/issues/608
Fixes https://github.com/scalecube/scalecube-services/pull/616

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/transport/api/ServerTransport.java
+++ b/services-api/src/main/java/io/scalecube/services/transport/api/ServerTransport.java
@@ -17,11 +17,10 @@ public interface ServerTransport {
   /**
    * Starts a server transport.
    *
-   * @param port listen port (can be {@code 0})
    * @param methodRegistry service method registry
    * @return bound server address
    */
-  Mono<ServerTransport> bind(int port, ServiceMethodRegistry methodRegistry);
+  Mono<ServerTransport> bind(ServiceMethodRegistry methodRegistry);
 
   /**
    * Stops server transport.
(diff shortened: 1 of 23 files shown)


eutkin:
technically, I had pulled similar to the first, but I have no N.I.H.-syndrome and I'm not against other visions :). @artem-v  Can you explain in more detail why you should remove the port from the bind?

In [#616](https://github.com/scalecube/scalecube-services/pull/616), I separated the tcp netty transport from rsocket. This can be useful if the client wants to use a web socket with rsocket.

artem-v:
@eutkin 
In #616  amount of changes doesn't reflect what's written in its motivation description. That's why it's hard to judge did you pulled similar to this one or not. From motivation description I understood u want to customize tcpServer and/or  tcpClient  of  the RSocketServiceTransport. So here u go. 


> Can you explain in more detail why you should remove the port from the bind?

There's issue for this https://github.com/scalecube/scalecube-services/issues/608. BTW even having `ServiceMethodRegistry` in `bind()` is debatable.

eutkin:
> @eutkin
> In #616 amount of changes doesn't reflect what's written in its motivation description. That's why it's hard to judge did you pulled similar to this one or not. From motivation description I understood u want to customize tcpServer and/or tcpClient of the RSocketServiceTransport. So here u go.
> 
> > Can you explain in more detail why you should remove the port from the bind?
> 
> There's issue for this #608. BTW even having `ServiceMethodRegistry` in `bind()` is debatable.

In this issue, you wrote what needs to be done, but not why

artem-v:
In https://github.com/scalecube/scalecube-services/issues/608  we want to have less coupling on addressing: ports/hosts and etc. It should be irrelevant (on interface level) on what port/host  i start server transport, what's relevant that it starts.

ronenhamias (approved):

PR #620, opened 2019-07-01
Move ServiceCall to api module.

snripa:
Allow to pass some fixed address to ServiceCall and let it be used instead of request routing.
As a consequence - it's up to Microservices to provide default router implementation (round-robin - as it was before)


services/src/main/java/io/scalecube/services/MicroservicesInjection.java
```diff
@@ -0,0 +1,102 @@
...
+                            })
+                        .toArray();
+                afterConstructMethod.invoke(targetInstance, parameters);
+              } catch (Exception ex) {
+                throw new RuntimeException(ex);
+              }
+            });
+  }
+
+  private static boolean isService(Class<?> type) {
```
ronenhamias: why not to use inject class for this?

services/src/main/java/io/scalecube/services/MicroservicesInjection.java
```diff
@@ -0,0 +1,102 @@
...
+    return type.isAnnotationPresent(Service.class);
+  }
+
+  private static void setField(Field field, Object object, Object value) {
```
ronenhamias: why not to use inject class for this?

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -5,16 +5,12 @@
 import static io.scalecube.services.CommunicationMode.REQUEST_RESPONSE;
```
artem-v, line 5: Please  don't  re-arrange functions. It's hard to review what was changed. Please revert back methods aligment as it was.

artem-v (changes requested):

artem-v (approved):

ronenhamias (approved):

PR #624, opened 2019-07-05
Design auth support

dmytro-lazebnyi:


services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -62,23 +65,48 @@ public static Type parameterizedReturnType(Method method) {
...
     }
   }
 
+  /**
+   * Return index of parameter which is defined as request.
+   *
+   * @param method method
+   * @return index of parameter
+   */
+  public static Integer requestParameterIndex(Method method) {
```
aharonha: `OptionalInt` can be nicer

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -270,7 +297,14 @@ public static void validateMethodOrThrow(Method method) {
       throw new UnsupportedOperationException("Service method return type can be Publisher only");
     }
     if (method.getParameterCount() > 1) {
-      throw new UnsupportedOperationException("Service method can accept 0 or 1 parameters only");
+      int paramsCount = 0;
+
+      for (Parameter parameter : method.getParameters()) {
```
aharonha: why don't you do an index loop?

services-api/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -364,31 +375,34 @@ public ServiceCall requestReleaser(Consumer<Object> requestReleaser) {
                 return check.get(); // toString, hashCode was invoked.
               }
 
+              Integer requestParameterIndex = Reflect.requestParameterIndex(method);
```
aharonha: please use `OptionalInt`

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
```diff
@@ -31,13 +31,15 @@
   public static final String HEADER_DATA_FORMAT = "_data_format";
 
   private Map<String, String> headers = new HashMap<>(1);
+  private Object credentials;
```
aharonha: Let's create a base class for it. I see it as misuse of `Object`.

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
```diff
@@ -132,10 +119,26 @@ public String dataFormat() {
...
   }
 
+  /**
+   * Returns credentials.
+   *
+   * @param <T> credentials type
+   * @return credentials
+   */
+  public <T> T credentials() {
+    // noinspection unchecked
```
aharonha: it's `noinspection unchecked` because it's an object an not a specific class. try to change it

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -106,6 +141,25 @@ public ServiceMethodInvoker(
...
 
+  @SuppressWarnings("unchecked")
+  private Mono<Object> authenticate(ServiceMessage message) {
+    return Mono.defer(
+        () -> {
+          if (!methodInfo.isAuth()) {
+            return Mono.empty();
+          }
+          Object credentials = message.credentials();
+          if (credentials == null) {
```
segabriel: It's redundant check, IMHO we shouldn't check it, we should just call `authenticatorImpl` and let it make a decision.

services-api/src/main/java/io/scalecube/services/auth/Authenticator.java
```diff
@@ -0,0 +1,20 @@
...
+
+import reactor.core.publisher.Mono;
+
+/**
+ * Authenticator.
+ *
+ * @param <C> credentials type
+ * @param <P> principal type
+ */
+public interface Authenticator<C, P> {
```
aharonha: If you create a class for credentials please bind it like `C extends Credentials`

services/src/test/java/io/scalecube/services/ServiceAuthLocalTest.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final Duration TIMEOUT = Duration.ofSeconds(3);
+
+  private static final String CREDENTIALS = "valid_credentials";
+  private static final Authenticator<String, UserProfile> authenticator =
+      credentials -> {
+        if (CREDENTIALS.equals(credentials)) {
+          return Mono.just(new UserProfile("Alice", "ADMIN"));
+        }
+
+        return Mono.error(new UnauthorizedException("Authentication failed"));
```
aharonha, line 45: `UnauthorizedException`? its authentication, not authorization.

services/src/test/java/io/scalecube/services/ServiceAuthRemoteTest.java
```diff
@@ -0,0 +1,214 @@
...
+import java.time.Duration;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
+final class ServiceAuthRemoteTest {
```
aharonha, line 26: 1. why is it final?
2. Can you use fixtures (one for local, one for remote)?

aharonha (changes requested):
see inline comments

services-api/src/main/java/io/scalecube/services/auth/Authenticator.java
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Returns principal by given credentials.
+   *
+   * @param credentials credentials
+   * @return async result with obtained principal
+   */
+  Mono<P> authenticate(C credentials);
```
artem-v: I think `C`  must be of type `Message` . `Message`  contains `headers` and they come handy along with `credentials` itself.

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -80,19 +90,44 @@ public ServiceMethodInvoker(
...
+              if (first.hasValue()) {
+                return authenticate(first.get())
+                    .flatMapMany(
+                        principal ->
+                            messages
+                                .map(this::toRequest)
+                                .transform(request -> invoke(request, principal)))
+                    .map(this::toResponse);
+              } else {
+                return messages
```
artem-v: How this `else`  block is possible?

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/Reflect.java
artem-v: Can we do it without `method.getParameters()` in the critical path ?

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -270,7 +297,14 @@ public static void validateMethodOrThrow(Method method) {
...
+        if (!parameter.isAnnotationPresent(Principal.class) && ++paramsCount > 1) {
```
artem-v: Can we do it without `arameter.isAnnotationPresent(Principal.class)`  in the critical path ?

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/Reflect.java
aharonha: we should cache this (as method signature is *Hard Coded*)

services-api/src/main/java/io/scalecube/services/Reflect.java
aharonha: we should cache this (as method signature is *Hard Coded*)

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
dmytro-lazebnyi: Removed.

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: I changed implementation, so `requestParameterIndex` does not exist now.

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: This part of the code doesn't exist anymore.

services-api/src/main/java/io/scalecube/services/ServiceCall.java
dmytro-lazebnyi: This part of the code doesn't exist anymore.

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
dmytro-lazebnyi: Credentials are part of headers now.

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
dmytro-lazebnyi: This part of the code doesn't exist anymore since credentials are part of headers.

services-api/src/main/java/io/scalecube/services/auth/Authenticator.java
dmytro-lazebnyi: Now `Authenticator` accepts `ServiceMessage` and is responsible to retrieve credentials from headers.

services/src/test/java/io/scalecube/services/ServiceAuthLocalTest.java
dmytro-lazebnyi: `UnathorizedException` is one of default `scalecube-service` exceptions.
Those exceptions have similar semantics to HTTP error codes (also, `errorType` of exception has exactly the same value as HTTP error code).

In HTTP:
- if authentication fails `401 Unauthorized` should be returned (see https://tools.ietf.org/html/rfc7235#section-3.1).
- if authorization fails `403 Forbidden` should be returned (see https://tools.ietf.org/html/rfc7231#section-6.5.3)

If you would like to introduce a new type of exception(s), please discuss it with @artem-v.

services/src/test/java/io/scalecube/services/ServiceAuthRemoteTest.java
dmytro-lazebnyi: 1. I prefer to mark every class as final unless I expect it can be inherited (with high probability). In this situation, I don't see a necessity to mark the test class as non-final.

2. Currently `scalecube-services` doesn't use `scalecube-test-utils` for tests at all. Maybe it is a good idea to use it, because there are already several tests for the local and remote environments, but I don't want to lose focus it this PR by adding more test-related code (adding an additional dependency and creating fixtures). If you see it can be useful, please create a separate issue.

services-api/src/main/java/io/scalecube/services/auth/Authenticator.java
dmytro-lazebnyi: Changed to use `ServiceMessage` instead of `C`.

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
dmytro-lazebnyi: Removed.

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: Implementation is changed a bit, but still contains `method.getParameters`.

Anyway, `validateMethodOrThrow` is not the critical path, since it is called only on service registration.

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: Implementation is changed a bit, but still contains `parameter.isAnnotationPresent(Principal.class)`.

Anyway, `validateMethodOrThrow` is not the critical path, since it is called only on service registration.

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -269,8 +275,22 @@ public static void validateMethodOrThrow(Method method) {
     } else if (!Publisher.class.isAssignableFrom(returnType)) {
       throw new UnsupportedOperationException("Service method return type can be Publisher only");
     }
-    if (method.getParameterCount() > 1) {
-      throw new UnsupportedOperationException("Service method can accept 0 or 1 parameters only");
+    if (method.getParameterCount() == 2) {
```
ronenhamias: this block can be improved... 
maybe consider to make verification method  since reading all these if is off context to the logical flow
something like
verifyPrincipelParam ...????

dmytro-lazebnyi: Fixed in https://github.com/scalecube/scalecube-services/pull/624/commits/824e427df1b98fe1640efbfe321e1471c706e37f

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -106,6 +122,38 @@ public ServiceMethodInvoker(
...
 
+  private Object[] prepareArguments(Object request, Object principal) {
+    Object[] arguments = new Object[methodInfo.parameterCount()];
+    Object principalArg = principal.equals(NO_PRINCIPAL) ? null : principal;
+
+    if (methodInfo.requestType() != Void.TYPE) {
+      arguments[0] = request;
+    } else {
+      arguments[0] = principalArg;
+    }
```
artem-v, line 135: Couldn't comprehend 129-133 (very confusing), until seen the changes in Reflect. Looks like u are using `Void.TYPE` to denote a fact that there're no 'normal' service method parameters. 

In other words method like following is _void_:  
` String hello(@Principal p) `
and these ones are not _void_: 
`String hello(String name, @Principal p)`
`String hello(String name)`

Correct?

services-api/src/main/java/io/scalecube/services/exceptions/UnauthorizedException.java
```diff
@@ -7,4 +7,8 @@
   public UnauthorizedException(int errorCode, String message) {
     super(errorCode, message);
   }
+
+  public UnauthorizedException(String message) {
+    super(ERROR_TYPE, message);
+  }
```
artem-v, line 13: There's little mess in http around terminologies   forbidden/unauthorized and authentication/authorization.  Plz take a look here:  https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses. 
So  term _forbidden_ stands for _authroization error_, and _unauthorized_  is for _authentication error_. Looks like we need  new class --  `ForbiddenException`.

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -324,4 +349,9 @@ public static void setField(Field field, Object object, Object value)
   public static boolean isService(Class<?> type) {
     return type.isAnnotationPresent(Service.class);
   }
+
+  public static boolean isAuth(Method method) {
+    return method.isAnnotationPresent(Auth.class)
+        || method.getDeclaringClass().isAnnotationPresent(Auth.class);
+  }
```
artem-v, line 365: `@Auth` on class or method turns a method (or all methods if on class) to _auth_ one. Ok.
But can a method (yes, one method)  be _auth_ one without defining `@Auth` annotation? For example client didn't put auth annotation but put `@Principal` on a parameter; Formally he didn't put auth annotation so method is not _auth_ one, but he has a principal  in signature, so he wants to be auth. Or put it otherwise -- if `@Principal` is defined and `@Auth` is not then we don't support such method definition.

services-api/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -45,6 +45,7 @@
       req -> {
         // no-op
       };
+  private Map<String, String> credentials;
```
artem-v: Plz assign this map to some `Collections.emptyMap()`.  Having this would simply remove the need in changes in ServiceMessage  (javadoc + `if` ccondition): 

```
    /**
     * Adds to service message each element of given <code>headers</code>.
     *
     * @param headers given headers
     * @return self
     */
    public Builder headers(Map<String, String> headers) {
      if (headers != null) {
        ...
      }
      return this;
    }
```

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -106,6 +122,38 @@ public ServiceMethodInvoker(
...
+        () -> {
+          if (!methodInfo.isAuth()) {
+            return Mono.empty();
+          }
+          if (authenticator == null) {
+            throw new UnauthorizedException("Authenticator not found");
+          }
+          return authenticator
+              .authenticate(message)
+              .onErrorMap(th -> new UnauthorizedException(th.getMessage()));
```
artem-v: Not all exceptions would have message. Maybe better to pass `th` itself?

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -40,12 +45,14 @@ public ServiceMethodInvoker(
...
       ServiceProviderErrorMapper errorMapper,
-      ServiceMessageDataDecoder dataDecoder) {
+      ServiceMessageDataDecoder dataDecoder,
+      Authenticator authenticator) {
     this.method = method;
     this.service = service;
     this.methodInfo = methodInfo;
     this.errorMapper = errorMapper;
     this.dataDecoder = dataDecoder;
+    this.authenticator = authenticator;
```
artem-v: This line complains with _Unchecked assignment_, plz fix.

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
artem-v, line 57: Unchecked assignment? Idea  complains.

services/src/test/java/io/scalecube/services/sut/security/SecuredServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Override
+  public Mono<String> helloWithRequestAndPrincipal(String name, UserProfile user) {
+    authorize(user);
+    return Mono.just("Hello, " + name + " and " + user.name());
+  }
+
+  private void authorize(UserProfile user) {
+    if (!user.role().equals("ADMIN")) {
+      throw new UnauthorizedException("Unathorized");
```
artem-v: IMHO here must be `ForbiddenException`

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
artem-v: @segabriel 
Suppose as an authenticator implementor now I have to deal with `null` as `credentials` in my function. Ok. What alternative to throwing Exception do I have?

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
dmytro-lazebnyi: Yep. Correct.

services-api/src/main/java/io/scalecube/services/Reflect.java
aharonha: One can put `@Auth` on method just to allow this method to be verified with the `Authenticator` and no further action (e.g. principal verification / access control) is needed.
If someone puts `@Principal` on a parameter without `@Auth` (neither method nor class) **I think** it should imply `@Auth` on that method.

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -62,6 +65,10 @@ public static Type parameterizedReturnType(Method method) {
       if (method.isAnnotationPresent(RequestType.class)) {
         return method.getAnnotation(RequestType.class).value();
       } else {
+        if (method.getParameters()[0].isAnnotationPresent(Principal.class)) {
+          return Void.TYPE;
```
aharonha, line 69: ```if (method.getParameterTypes().length > 0) {``` can also mean 2 parameters.

services-api/src/main/java/io/scalecube/services/exceptions/UnauthorizedException.java
dmytro-lazebnyi: Added https://github.com/scalecube/scalecube-services/pull/624/commits/98eb543efb827ec0afa068c24e5b2216696574b4

services-api/src/main/java/io/scalecube/services/Reflect.java
```diff
@@ -269,8 +275,36 @@ public static void validateMethodOrThrow(Method method) {
...
+          throw new UnsupportedOperationException(
+              "@Principal can be used only for parameter of @Auth method");
+        }
+      }
+    }
+
+    if (method.getParameterCount() == 2) {
+      if (parameters[0].isAnnotationPresent(Principal.class)) {
+        throw new UnsupportedOperationException(
+            "@Principal cannot be the first parameter if parameters count equals 2");
```
aharonha:
```suggestion
            "@Principal cannot be the first parameter on method with two parameters");
```

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: Added additional validation https://github.com/scalecube/scalecube-services/pull/624/commits/b8b37531fd0371cd1708f10b6b6687c6f1f18b0a

services-api/src/main/java/io/scalecube/services/ServiceCall.java
dmytro-lazebnyi: Done https://github.com/scalecube/scalecube-services/pull/624/commits/72c78acfe6d73bb7d213c3cb2b6d0435a5715251

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
dmytro-lazebnyi: Suppressed https://github.com/scalecube/scalecube-services/pull/624/commits/59862a44bfb2bf0c15108234da66a4c3440ee0dc

services/src/test/java/io/scalecube/services/sut/security/SecuredServiceImpl.java
dmytro-lazebnyi: Fixed https://github.com/scalecube/scalecube-services/pull/624/commits/98eb543efb827ec0afa068c24e5b2216696574b4

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
aharonha: it's because of `<Object>` please bind to a new `Principal` interface

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
```diff
@@ -106,6 +123,38 @@ public ServiceMethodInvoker(
...
+        () -> {
+          if (!methodInfo.isAuth()) {
+            return Mono.empty();
+          }
+          if (authenticator == null) {
+            throw new UnauthorizedException("Authenticator not found");
+          }
+          return authenticator
+              .authenticate(message)
+              .onErrorMap(th -> new UnauthorizedException(th.getMessage()));
```
segabriel: So in that way, do we lose `errorCode` the thrown exception?

dmytro-lazebnyi: Fixed https://github.com/scalecube/scalecube-services/pull/624/commits/a02e7be740b3671bff33430f668c35d6969bebbd

services-api/src/main/java/io/scalecube/services/Reflect.java
artem-v: @aharonha 
What you say makes sense. But let's not do it from start, unless we face this problem down the road: when somebody start to use this and start to complain on us   _hey why putting `@Principal` is not enough?_

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
dmytro-lazebnyi: Done https://github.com/scalecube/scalecube-services/pull/624/commits/a02e7be740b3671bff33430f668c35d6969bebbd

artem-v (approved):

segabriel (approved):

aharonha (approved):

services-api/src/main/java/io/scalecube/services/methods/ServiceMethodInvoker.java
aharonha: you can also change the member to be `Authenticator<? extends Object>`

services-api/src/main/java/io/scalecube/services/Reflect.java
dmytro-lazebnyi: Yep.

PR #626, opened 2019-07-06
Integrate with cluster 2.4.x

artem-v:
This is draft PR. Aim is to collect objections/propositions. And to actually have a look at integration with services before making release of 2.4.0 scalecube-cluster.


services-examples-parent/services-examples-runner/src/main/java/io/scalecube/services/examples/ExamplesRunner.java
```diff
@@ -81,8 +81,8 @@ private static ServiceDiscovery serviceDiscovery(ServiceEndpoint serviceEndpoint
     return new ScalecubeServiceDiscovery(serviceEndpoint)
         .options(
             opts ->
-                opts.seedMembers(config.seedAddresses())
-                    .port(config.discoveryPort())
+                opts.membership(cfg1 -> cfg1.seedMembers(config.seedAddresses()))
```
ronenhamias, line 84: this api change is less clean... less expect-able.

artem-v: With this api change cluster configuration comes in groups: MembershipConfig, TransportConifg, GossipConfig and ClusterConfig  itself. It was very inconvenient when settings from all sub-components were living in global CluserConfig. Now with separation of configuration options it's clear what properties has gossip, what properties  has FailureDetector, and so on.

PR #630, opened 2019-07-15
Enrich ServiceDiscoveryEvent `toString`

segabriel:


services-api/src/main/java/io/scalecube/services/discovery/api/ServiceDiscoveryEvent.java
```diff
@@ -142,7 +142,7 @@ public String toString() {
             .orElse(null)
         + ", serviceEndpoints="
         + serviceEndpoints.stream() //
-            .map(ServiceEndpoint::id)
+            .map(se -> se.id() + "( " + se.address() + ")")
```
artem-v: Must be in format ID@ADDRESS

artem-v (changes requested):

artem-v (approved):

PR #633, opened 2019-07-17
`RequestType` and `ResponseType` support

dmytro-lazebnyi:


segabriel (approved):
:+1:

artem-v (changes requested):
*  If `ServiceMessage` is defined as return type and  no `@ResponseType` is defined -- not allowed. 
* If `ServiceMessage` is defined as request type and no `@RequestType` is defined -- not allowed.

PR #686, opened 2019-10-27
Add test on discovery

artem-v:


artem-v:
@segabriel 
FYI 
I have added two new tests to check your assumptions about problems when using some of the group member as a seed: `testServiceGroupDiscoveryAsSeed`, `testServiceGroupDiscoveryAsSeedOnBehalfOfNonGroupMember`. Found no problems.

PR #710, opened 2020-02-18
WIP: [ISSUE-708] Extract interface from Microservices

eutkin:
Modification sketches for DI extraction to a separate module.

 Sample plan:
1. Create new interface for Microservices.
2. Create new interface for ServiceProvider (`[Microservices, Mono<Collection<ServiceInfo>>]`) for lazy services initialization. For example, Spring: 
	```java
	microservices -> {
	  var appContext = new GenericApplicationContext();
	  // setup beans
	  return Mono.defer(() -> 
				appContext.getBeansByAnnotation(Service.class));
	}
	```
3. Replace service instance to service instance supplier (for lazy) and service interface with `@Service` (as `Class<?>`) in `ServiceInfo`.
4. Change method register, method invoker (linked change set of ServiceInfo)
5. ???
6. Profit


services-api/src/main/java/io/scalecube/services/IMicroservices.java
```diff
@@ -0,0 +1,13 @@
+package io.scalecube.services;
+
+import io.scalecube.net.Address;
+import io.scalecube.services.discovery.api.ServiceDiscovery;
+
+public interface IMicroservices {
```
ronenhamias: how about to rename current microservices class to ScaleCube

so:

`ScaleCube implements Microservices`

since we dont use `I` as prefix to interfaces.
further more if this class becomes an interface then it should be moved to services.api

PR #711, opened 2020-03-11
Extract DI to separate module

### LINKED ISSUE #708: Separating non-core functionality into separate modules
## Problem 
ScaleCube Services has supporting subsystems that allow it to be self-contained. For now, I want to highlight two subsystems:
- **subsystem IoC (DI)**: `@Inject`, etc;
- **subsystem Security**: `@Principal`, `Authenticator`, etc.

These subsystems are hardcoded into the project, which prevents ScaleCube Services from integrating with third-party implementations of IoC/Security frameworks: Spring IoC, Guice, Spring Security, Okta, etc. 

Problems that prevent integration:
- duplication of functionality;
- conflict of namespaces of different frameworks: the same for annotations `@Principal`, `@Inject` ;
- cyclical dependence in ScaleCube IoC between instance service and instance `Microservices`;
- a mechanism for solving the type of parameter corresponding to the user's profile in the service method.

## Solution
I propose the following algorithm:

1. to separate the subsystems described above into separate modules;
2. to replace them with SPI interfaces;
3. to connect modules from item 1 as default implementations
4. write modules implementation using popular frameworks (I can take over completely).

## Motivation
I faced this problem when integrating the project with the Spring Framework. Integration is possible, but it looks redundant and crooked.

eutkin:
I've made life cycle management of scalecube services a separate subsystem: ServicesProvider. 

Pro:
 - The service lifecycle management logic is now concentrated in one class (ServicesProvider);
- You can replace the implementation with another IoC container (Spring/Guice/etc).
- The API is backwards compatible

idea

ServicesProvider - interface for IoC container that performs several functions:
- stores all instances of services;
- manages the life cycle of stored services (creation -> initialization -> finalization);
- gives meta information about stored services (type and tags).

Scale Cube for DI proprietary mechanisms are collected in one class, in ServicesProvider implementation (see ScaleCubeServicesProvider).

The Spring IoC implementation is also given for the example.


services-api/src/test/java/io/scalecube/services/methods/ServiceMethodInvokerTest.java
```diff
@@ -152,8 +160,10 @@ void testInvokeOneWhenThrowException() throws Exception {
             false,
             AUTH);
 
+    // FIXME: 19.02.2020
```
ronenhamias: whats needs to be fixed here?

segabriel:
Please fix imports, I came across with the same problem recently and the best solution is https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides

> The import ordering is not handled by this plugin, unfortunately. To fix the import order, download the IntelliJ Java Google Style file and import it into File→Settings→Editor→Code Style.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Default {@link ServicesProvider}.
+ */
+public class ScaleCubeServicesProvider implements ServicesProvider {
+
+  private final Function<Microservices, Collection<ServiceInfo>> serviceFactory;
+
+  // lazy init
+  private Collection<ServiceInfo> services;
```
segabriel: Can we avoid lazy init? Probably, need to apply thread-safe stuff for it.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,101 @@
...
+      Microservices microservices) {
+    this.services =
+        this.serviceFactory.apply(microservices).stream()
+            .map(service -> Injector.inject(microservices, service))
+            .collect(Collectors.toList());
+    List<ServiceDefinition> definitions =
+        this.services.stream()
+            .map(serviceInfo -> new ServiceDefinition(serviceInfo.type(), serviceInfo.tags()))
+            .collect(Collectors.toList());
+    return Mono.just(definitions);
```
segabriel: If the method returns `Mono` an invoker assumes that work will be started only after subscribing, so wrap it and similar things via `Mono.fromCallable/defer`.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,101 @@
...
+   * {@inheritDoc}
+   *
+   * <p>Use {@link io.scalecube.services.annotations.AfterConstruct} for initialization service's
+   * instance.
+   */
+  @Override
+  public Mono<? extends Collection<ServiceInfo>> provideService(Microservices microservices) {
+    return Mono.defer(
+        () ->
+            Mono.fromCallable(
```
segabriel: `Mono.fromCallable` is enough here, no need extra `Mono.defer` wrapping

services/src/main/java/io/scalecube/services/ScaleCube.java
```diff
@@ -113,14 +122,14 @@
  *
  * }</pre>
  */
-public final class Microservices {
+public final class ScaleCube implements Microservices {
```
segabriel: Maybe I missed name voting, but `Scalecube` looks better, with no camelCase no spaces between.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
segabriel:
```suggestion
    return Mono.fromCallable(
        () ->
            this.serviceFactory.apply(microservices).stream()
                .map(service -> Injector.inject(microservices, service))
                .map(serviceInfo -> new ServiceDefinition(serviceInfo.type(), serviceInfo.tags()))
                .collect(Collectors.toList()));
```

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,101 @@
...
+                () ->
+                    this.services.stream()
+                        .map(service -> Injector.processAfterConstruct(microservices, service))
+                        .collect(Collectors.toList())));
```
segabriel:
```suggestion
    return Mono.fromCallable(
        () ->
            this.serviceFactory.apply(microservices).stream()
                .map(service -> Injector.processAfterConstruct(microservices, service))
                .collect(Collectors.toList()));
```

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * {@inheritDoc}
+   *
+   * <p>Use {@link io.scalecube.services.annotations.BeforeDestroy} for finilization service's
+   * instance.
+   */
+  @Override
+  public Mono<Microservices> shutDown(Microservices microservices) {
+    this.services.forEach(service -> Injector.processBeforeDestroy(microservices, service));
+    return Mono.just(microservices);
```
segabriel:
```suggestion
    return Mono.fromRunnable(
            () ->
                this.serviceFactory
                    .apply(microservices)
                    .forEach(service -> Injector.processBeforeDestroy(microservices, service)))
        .thenReturn(microservices);
```

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
segabriel:
```suggestion
```

services/src/main/java/io/scalecube/services/ScaleCube.java
ronenhamias: +1 for Scalecube no camelcase!

services/src/main/java/io/scalecube/services/ScaleCube.java
```diff
@@ -331,40 +344,89 @@ private void listenJvmShutdown() {
...
+                          .dataDecoderIfAbsent(this.dataDecoder)
+                          .errorMapperIfAbsent(this.errorMapper)
+                          .build())
+              .collect(
+                  Collectors.collectingAndThen(
+                      Collectors.toList(), services -> (ServiceProvider) call -> services));
+      this.serviceProviders.add(serviceProvider);
+      this.servicesProvider =
+          this.servicesProvider == null
+              ? ScaleCubeServicesProvider.create(this.serviceProviders)
```
segabriel: Please pass a new immutable collection.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
segabriel: See suggestions below.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
segabriel: See suggestions below.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
segabriel: See suggestions below.

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  public SpringServicesProvider(Class<?>... configurations) {
+    this.configurations = configurations;
+  }
+
+  @Override
+  public Mono<Collection<ServiceDefinition>> provideServiceDefinitions(
+      Microservices microservices) {
+
+    this.ctx = new AnnotationConfigApplicationContext();
```
segabriel: Move it to constructor and make it as final field.

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    discoveryBd.setLazyInit(true);
+    discoveryBd.setSynthetic(true);
+    discoveryBd.setInstanceSupplier(instanceSupplier);
+    discoveryBd.setSource(this);
+    return discoveryBd;
+  }
+
+  @Override
+  public Mono<? extends Collection<ServiceInfo>> provideService(Microservices microservices) {
+    return Mono.defer(
```
segabriel: Please use `Mono.fromCallbale` instead of it

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    List<ServiceDefinition> serviceDefinitions =
+        Stream.of(serviceBeanNames)
+            .map(serviceBeanName -> this.ctx.getBeanDefinition(serviceBeanName))
+            .map(BeanDefinition::getBeanClassName)
+            .filter(Objects::nonNull)
+            .map(className -> ClassUtils.resolveClassName(className, ctx.getClassLoader()))
+            .filter(type -> type.getAnnotation(Service.class) != null)
+            .map(type -> new ServiceDefinition(type, Collections.emptyMap()))
+            .collect(Collectors.toList());
+    return Mono.defer(() -> Mono.just(serviceDefinitions));
```
segabriel: Cool construction but useless, please wrap whole given method into `Mono.fromCallable` and just return `serviceDefinitions`.

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
```diff
@@ -0,0 +1,103 @@
...
+              this.ctx.getBeansWithAnnotation(Service.class).values().stream()
+                  .map(bean -> ServiceInfo.fromServiceInstance(bean).build())
+                  .collect(Collectors.toList());
+          return Mono.just(services);
+        });
+  }
+
+  @Override
+  public Mono<Microservices> shutDown(Microservices microservices) {
+    return Mono.defer(
```
segabriel: Unwrap it, `Mono.fromCallable` is enough

services-api/src/test/java/io/scalecube/services/methods/ServiceMethodInvokerTest.java
eutkin: Outdated changes. I used to think of doing a Supplier<*ServiceInstance* instead of a service instance in ServiceInfo to get a service instance lazy. Otherwise, you couldn't inject ServiceDiscovery because it hadn't yet been initialized when the service instance was created.

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
eutkin: Thank you for the remark. The original implementation was thread safe, but 
- this service was not intended to be made public api
- its use is limited to the inside of ScaleCube

So I decided not to overcomplicate the implementation

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
eutkin: This is only a temporary example for internal use. Before I merged this pull request, I was supposed to delete it. Since this is just a demonstration of how to use Spring IoC

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Default {@link ServicesProvider}. Thread-safe.
+ */
+public class ScaleCubeServicesProvider implements ServicesProvider {
+
+  private final Function<Microservices, Collection<ServiceInfo>> serviceFactory;
+
+  // lazy init
+  private AtomicReference<Collection<ServiceInfo>> services = new AtomicReference<>();
```
segabriel: Remove it at all, see suggestions

ronenhamias:
since .services(new MyService()) is deprecated
we need to update all the examples to how to use the current api correctly including the docs.

```java
// Construct a ScaleCube node which joins the cluster hosting the Greeting Service
    Scalecube ms =
        Scalecube.builder()
            .discovery(
                endpoint ->
                    new ScalecubeServiceDiscovery(endpoint)
                        .membership(cfg -> cfg.seedMembers(seedAddress)))
            .transport(RSocketServiceTransport::new)
            .services(new BidiGreetingImpl())
            .startAwait();
```

services-api/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,15 @@
+package io.scalecube.services;
+
+import io.scalecube.net.Address;
+import io.scalecube.services.discovery.api.ServiceDiscovery;
+
+public interface Microservices {
```
artem-v: Leave `Microservices` as it was. If u need references to some parts of `Microservices` during some phases in external pluggable 3party components => then expose those _microservices parts_ via `Microservices.Context` (inner `public`  interface, implementation is inner `private static`).

services-api/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -1,506 +1,507 @@
-package io.scalecube.services;
```
artem-v: What changed in `ServiceCall`? Why it's `+507` `-506`?

services-api/src/main/java/io/scalecube/services/ServiceInfo.java
```diff
@@ -85,26 +85,82 @@ private Builder(Object serviceInstance) {
...
       return this;
     }
 
+    /**
+     * Set up {@link ServiceProviderErrorMapper} if it hasn't been set up before.
+     *
+     * @param errorMapper error mapper.
+     * @return current builder's state.
+     */
+    public Builder errorMapperIfAbsent(ServiceProviderErrorMapper errorMapper) {
```
artem-v: Didn't get idea of those `.***IfAbsent`  methods. Why u need them? Irregardless of `servicesProvider` type u go through `.registerInMethodRegistry()` anyway which, by turn, sets corresponding dependencies.

services-api/src/main/java/io/scalecube/services/ServiceInfo.java
```diff
@@ -85,26 +85,82 @@ private Builder(Object serviceInstance) {
...
+      return this;
+    }
+
+    /**
+     * Set up {@link ServiceMessageDataDecoder} if it hasn't been set up before.
+     *
+     * @param dataDecoder data decoder.
+     * @return current builder's state.
+     */
+    public Builder dataDecoderIfAbsent(ServiceMessageDataDecoder dataDecoder) {
```
artem-v: Same here.

services-api/src/main/java/io/scalecube/services/ServiceProvider.java
```diff
@@ -2,7 +2,13 @@
 
 import java.util.Collection;
 
+/**
+ * Provide service instances.
+ *
+ * @deprecated use {@link ServicesProvider}
+ */
 @FunctionalInterface
+@Deprecated
```
artem-v, line 11: All is cool with this interface. It shouldn't be deprecated. I still want to use good-old approach with `.services(Object... ) / .services(ServiceInfo... )` or `services(ServiceProvider)`. Keep `deprecated` mentioned top level functions on  Microservices (or how u call it here _Scalecube_), but decalre them on the `ScaleCubeServicesProvider` (with the same fluent API style).

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
```diff
@@ -0,0 +1,35 @@
...
+import reactor.core.publisher.Mono;
+
+/** Manages the life cycle of all services registered with Scale Cube Services. */
+public interface ServicesProvider {
+
+  /**
+   * Provide service definitions.
+   *
+   * @param microservices scale cube instance.
+   * @return collection of service definitions - service type&amp;tags.
```
artem-v: `&amp;`? just put a word `and`.

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
```diff
@@ -0,0 +1,34 @@
+package io.scalecube.services;
+
+import java.util.Collection;
+import reactor.core.publisher.Mono;
+
+/** Manages the life cycle of all services registered with Scale Cube Services. */
+public interface ServicesProvider {
```
artem-v: Name it `ServiceFactory`. Too much _providers_ already.

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Provide service definitions.
+   *
+   * @param microservices scale cube instance.
+   * @return collection of service definitions - service type&amp;tags.
+   * @see ServiceDefinition
+   */
+  Mono<? extends Collection<ServiceDefinition>> provideServiceDefinitions(
```
artem-v: Remove `Mono` use good-old collection. Name it `getServiceDefinitions()`.

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
```diff
@@ -0,0 +1,34 @@
...
+      Microservices microservices);
+
+  /**
+   * Initialize instances of services.
+   *
+   * @param microservices scale cube instance.
+   * @return Completed Mono if initialization was successful for all services.
+   */
+  Mono<? extends Collection<ServiceInfo>> provideService(Microservices microservices);
```
artem-v: Name it `.initServices()` or `.initializeServices()`.

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Finalization of service instances.
+   *
+   * @param microservices scale cube instance.
+   * @return completed Mono if finalization was successful for all services.
+   */
+  Mono<Microservices> shutDown(Microservices microservices);
```
artem-v: Name it `.shutdownServices()`.

services-api/src/main/java/io/scalecube/services/routing/Routers.java
```diff
@@ -1,35 +1,35 @@
-package io.scalecube.services.routing;
```
artem-v: What has changed here?

services-benchmarks/src/main/java/io/scalecube/services/benchmarks/transport/codec/SmCodecBenchmarkState.java
```diff
@@ -1,183 +1,183 @@
-package io.scalecube.services.benchmarks.transport.codec;
```
artem-v: What changed here?

artem-v (changes requested):

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
```diff
@@ -0,0 +1,103 @@
...
+        () ->
+            Mono.fromCallable(
+                () -> {
+                  this.ctx.stop();
```
artem-v: Don't u need to check for a `null` here?

services-examples-parent/scalecube-examples-spring/src/main/java/io/scalecube/spring/SpringServicesProvider.java
artem-v: > This is only a temporary example for internal use. Before I merged this pull request, I was supposed to delete it. Since this is just a demonstration of how to use Spring IoC

Whole idea is to create an example of **working** custom DI mechanism. Not a partial support not a half-working solution, but holistic one. So if u're going to delete exactly this one, then create another ones to show: how to use it with a simple standalone spring, how to use in the context of springboot.

services-api/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -0,0 +1,15 @@
...
+
+  String id();
+
+  ServiceCall call();
+
+  Address serviceAddress();
```
artem-v: In new `Microservices.Context` call plz `address()`. We don't expose any other address => no need to designate specifically that this is _service_ address.

artem-v (dismissed):

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
artem-v: > Remove it at all, see suggestions

Where those suggestions are?

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
eutkin: Nope. ServiceDefinition maybe blocking operation. Example. ScalecubeServicesProvider can only give out the types of services after their initialization.

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
eutkin: ok

services-api/src/main/java/io/scalecube/services/ServicesProvider.java
eutkin: ok

services-api/src/main/java/io/scalecube/services/Microservices.java
eutkin: Sounds great. But there's one thing. ServiceFactory (ex ServicesProvider) is in the service-api module, which is quite reasonable. And Scalecube (ex Microservices) is in the implementation module, spring-services. I do not think it is worth creating service-api => scalecube-services.

Moreover, the Microservices interface fully defines the functionality of the old Microservices (now Scalecube), not a part of it. 

The only thing that I would think about is choosing names because of backward compatibility. Keep the old name for the Microservices implementation, and rename the interface

services-api/src/main/java/io/scalecube/services/ServiceInfo.java
eutkin: because this looks pretty, see:

```java
.errorMapper(Optional.ofNullable(serviceInfo.authenticator()).orElse(this.errorMapper))
```
vs

```java
.errorMapperIfAbsent(this.defaultErrorMapper)
```

similar to Map#putIfAbsent.

services-api/src/main/java/io/scalecube/services/ServiceProvider.java
eutkin: Sometimes you have to throw away the old one and move on :)  And seriously, this API is deprecated for the following reason: we delegate tasks to store and manage service life cycle to ServiceFactory (ex ServicesProvider) implementations. Scalecube (Microservices) should not perform such tasks, imho. Especially if we allocate a separate component that solves these tasks, I see no sense in leaving the old API, it's better to transfer these methods to the new one. As an option, you can add `registerDefinition(ServiceInfo serviceInfo)` and `<T> registerDefinition(Supplier<T> instanceSupplier, Class<T> type, Tag... tags)` methods to ServiceFactory (ex ServiceProducer). 

In the end.
- I find it wrong to leave the API `.services(ServiceInfo... services)` in Scalecube (Microservices);
- Take this API into the interface and implement it.

P.S. During development, I added the registration of additional services through `register(ServiceInfo service)` but decided to remove it because it violated the immunity of the ServiceFactory (ServicesProvider).

services/src/main/java/io/scalecube/services/inject/ScaleCubeServicesProvider.java
eutkin: @segabriel I'm sorry, I didn't really know exactly how to do this.

services-api/src/main/java/io/scalecube/services/ServiceProvider.java
ronenhamias: its ok to deprecate  .services(ServiceInfo... services) but we cant just do that without discussing the public api.
services is used in all the examples and documentations and we need to direct the users as part of this pull request to a proper use of the api with examples / readme / website and basically everywhere.

plust we are changing this method we also need to discuss the public api from end user point of view. we need to review it to make sure its clean and easy to use.

eutkin: @ronenhamias So api is not removed, but @Deprecated :)

services-api/src/main/java/io/scalecube/services/Microservices.java
artem-v: >  I do not think it is worth creating service-api => scalecube-services.

1. Put `Microservices.Context` into `services-api`.
2. Make dependency on it from `Microservices` (so u would still have normal dependency `sc-services => sc-services-api`).
3. Leave `Microservices` as it was.

services-api/src/main/java/io/scalecube/services/ServiceInfo.java
artem-v: > because this looks pretty

It's subjective is it pretty or not. If this is the only argument then revert those `.***IfAbsent` changes.

services-api/src/main/java/io/scalecube/services/ServiceProvider.java
artem-v: > we delegate tasks to store and manage service life cycle to ServiceFactory (ex ServicesProvider) implementations

Correct. We don't want `.services(Object... ) / .services(ServiceInfo... ) or services(ServiceProvider)` on `Microservices` anymore.  That's why:
>  Keep `@deprecated`  functions `.services(Object... ) / .services(ServiceInfo... ) or services(ServiceProvider)` on 

Next.
> As an option, you can add registerDefinition(ServiceIn ...

We don't need options. We need clear and well understood approach. Shall we discuss it here? No. Not in the scope of this PR. That's why: 
> Declare them on the `ScaleCubeServicesProvider` (with the same fluent API style).


Upon this one.
> I see no sense in leaving the old API, it's better to transfer these methods to the new one. 

Good. But not in this PR. In this PR we will keep them but just move to another place.

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,16 @@
+package io.scalecube.services;
```
ronenhamias, line 1: move this class to services.api package

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,16 @@
...
+public interface MicroservicesContext {
+
+  String id();
+
+  ServiceCall call();
+
+  Address serviceAddress();
+
+  ServiceDiscovery discovery();
+
```
artem-v, line 37: Remove this line. Add java docs.

services-api/src/main/java/io/scalecube/services/ServiceFactory.java
```diff
@@ -0,0 +1,36 @@
...
+public interface ServiceFactory {
+
+  /**
+   * Provide service definitions.
+   *
+   * @param microservices scale cube instance.
+   * @return collection of service definitions - service type and tags.
+   * @see ServiceDefinition
+   */
+  Mono<? extends Collection<ServiceDefinition>> getServiceDefinitions(
```
artem-v: Why it's a Mono here? What a usecase?

artem-v (changes requested):

artem-v (changes requested):
Create module with spring examples. And guice examples. It's not clear what we gain from all this story. It's clear what we loose, but it's not clear what we gain.

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,16 @@
+package io.scalecube.services;
+
+import io.scalecube.net.Address;
+import io.scalecube.services.discovery.api.ServiceDiscovery;
+
+public interface MicroservicesContext {
+
+  String id();
+
+  ServiceCall call();
```
artem-v: Let's rename it this legacy `.call()` to `.serviceCall()`.  @ronenhamias wdyt?

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,16 @@
...
+
+  Address serviceAddress();
+
+  ServiceDiscovery discovery();
```
artem-v: Let's rename it this legacy `.discovery()` to `.serviceDiscovery()`.  @ronenhamias wdyt?

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ServiceFactory.java
eutkin: Lazy init spring context?

eutkin:
> Create module with spring examples. And guice examples. It's not clear what we gain from all this story. It's clear what we loose, but it's not clear what we gain.

done

services-api/src/main/java/io/scalecube/services/RemoteService.java
```diff
@@ -0,0 +1,4 @@
+package io.scalecube.services;
+
+/** Interface-marker for Proxy of remote services. */
+public interface RemoteService {}
```
segabriel: Seems like workaround with bad name (service call can be local).
And I see you use it in factories like the following:

```java
    @Override
    public Mono<? extends Collection<ServiceInfo>> initializeServices(
        MicroservicesContext microservices) {
      return Mono.fromCallable(
          () -> {
            this.context.start();
            return this.context.getBeansWithAnnotation(Service.class).values().stream()
                .filter(
                    Predicate.not(bean -> RemoteService.class.isAssignableFrom(bean.getClass())))
                .map(bean -> ServiceInfo.fromServiceInstance(bean).build())
          });
    }
```

Could we replace 

```java
RemoteService.class.isAssignableFrom(bean.getClass())
```

to use smth like?: 

```java
Reflect.isService(bean.getClass())
```

services-api/src/main/java/io/scalecube/services/ServiceFactory.java
segabriel: @artem-v are right, we don't need to have Mono in this interface at all. If you need such delusion as lazy initialization you can provide it at level up.

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
```diff
@@ -12,7 +13,8 @@
 
 public final class GreetingServiceImpl implements GreetingService {
 
-  @Inject Microservices ms;
+  @Inject
+  MicroservicesContext ms;
```
segabriel, line 17: Can we revert it? Use Microservices instead of its context.

services/src/test/java/io/scalecube/services/ServiceRegistryTest.java
```diff
@@ -119,16 +120,18 @@ public void test_listen_to_discovery_events(MetadataCodec metadataCodec) {
...
                       .discovery(defServiceDiscovery(seedAddress, metadataCodec))
                       .transport(RSocketServiceTransport::new)
-                      .services(new GreetingServiceImpl())
+                      .serviceFactory(ScalecubeServiceFactory.fromInstances(new GreetingServiceImpl()))
                       .startAwait();
               cluster.add(ms2);
             })
         .assertNext(event -> assertEquals(ENDPOINT_ADDED, event.type()))
-        .then(() -> cluster.remove(2).shutdown().block(TIMEOUT))
+        .then(() -> cluster.remove(2).shutdown().block())
```
segabriel: Let's keep TIMEOUT here

services-api/src/main/java/io/scalecube/services/RemoteService.java
eutkin: We need way for detecting proxy for remote service and local service. 
`Reflect.isService(bean.getClass)` don't do it

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
eutkin: Why?

services-api/src/main/java/io/scalecube/services/RemoteService.java
segabriel: Assume, but you have in arguments `MicroservicesContext` instance and you can filter out something via its providing API.

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
segabriel: I'd like to have the `Microservices` instance. Will do it still work if I request `Microservices`? And if no then why?

eutkin: @artem-v help, it is your idea :) What is there in microservices that microservicesContext does not? The idea was to limit Microservices to an interface that only provides what you really need. I think the question is more philosophical.

services-api/src/main/java/io/scalecube/services/RemoteService.java
eutkin: MicroservicesContext cant resolve this problem. MicroservicesContext need for inject this in other services, no more.

eutkin: please, see picture in top this PR

segabriel: I'd like to remove this class.
Look the given context has `io.scalecube.services.MicroservicesContext#serviceDiscovery` and this one has `io.scalecube.services.discovery.api.ServiceDiscovery#serviceEndpoint` so you have the ability to understand what this instance provides (local call) and what not (remote call).

segabriel: Or let's just provide a new method which will give us them in `MicroservicesContext`.
like `Collection<Class> services()`

eutkin: Yep, ok, this is redundant. But, how do detect proxy and local services in guice? I don't know :)

segabriel: Please reread the above comments

ronenhamias:
+1

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
ronenhamias: i kind of agree with @segabriel  the idea of having microservices is to grant access to the service to lower levels apis of scalecube such as service call, discovery, membership

eutkin: this access gives exactly MicroservicesContext

services-api/src/main/java/io/scalecube/services/RemoteService.java
eutkin: fixed

services-api/src/main/java/io/scalecube/services/ServiceFactory.java
eutkin: What if we're reading from the disk?

eutkin: What are the arguments against?

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
eutkin: @ronenhamias @segabriel I have now split MicroservicesContext into MicroservicesContext and ExtendedMicroservicesContext because we cannot access ServiceDiscovery during the ServiceDefinition collection phase. 

I'm sure you think this is a controversial step, but it's done specifically to prevent NPEs when using ServiceDiscovery in custom ServiceFactory implementations.

Let me remind you, MicroservicesContext provides ServiceCall, network address of node and id of node. ExtendedMicroservicesContext additional provides ServiceDiscovery instance.

ronenhamias: well for me it looks complicated maybe we should think more about that.
microservices is not just about service call
else we could just inject service call...

just thinking....

artem-v: https://github.com/scalecube/scalecube-services/pull/711#discussion_r437977735

Yes
The purpose of `MicroservicesContext` is to hide `Microservices` from client code. So service developer can see only some view of functions/objects from Microservices, plus it will depend only on service-api module, which is somewhat cleaner than to pull entire services library.

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Context of Scale Cube node. Used in {@link ServiceFactory}.
+ *
+ * @see ServiceFactory
+ */
+public interface MicroservicesContext {
+
+  /**
+   * Id of Scale Cube node.
```
artem-v: Turn `Scale Cube` or `ScaleCube` to `Scalecube` evrey in the comments and variable names

services-api/src/main/java/io/scalecube/services/ExtendedMicroservicesContext.java
```diff
@@ -0,0 +1,21 @@
...
+import io.scalecube.net.Address;
+import io.scalecube.services.discovery.api.ServiceDiscovery;
+
+/**
+ * Context of Scale Cube node. Used in {@link ServiceFactory}. Compared to MicroservicesContext, it
+ * gives access to ServiceDiscovery.
+ *
+ * @see ServiceFactory
+ */
+public interface ExtendedMicroservicesContext extends MicroservicesContext {
```
artem-v: Let's have single one `MicroserviceContext` interface.

services-api/src/main/java/io/scalecube/services/ExtendedMicroservicesContext.java
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+   * Service discovery for services localed in other nodes.
+   *
+   * @return service discovery
+   * @see ServiceDiscovery
+   */
+  ServiceDiscovery serviceDiscovery();
```
artem-v: Move it to `MicroserviceContext`

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ServiceDefinition.java
```diff
@@ -0,0 +1,28 @@
+package io.scalecube.services;
+
+import java.util.Collections;
+import java.util.Map;
+
+/** Definition of service - type and tags. */
+public class ServiceDefinition {
```
artem-v, line 7: Call it `ServiceType`, service definition is too loud name for a tuple of Class+HashMap

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ExtendedMicroservicesContext.java
eutkin: ServiceFactory#getServiceDefinitions consume MicroserviceContext, but MicroserviceContext#serviceDiscovery return null

services-api/src/main/java/io/scalecube/services/ServiceDefinition.java
eutkin: I don't think that's a sufficient argument. I see it as a service definition, perhaps in the future this entity will contain more information

services-api/src/main/java/io/scalecube/services/ExtendedMicroservicesContext.java
artem-v: Until u read service-definitions discovery will not be ready. It's a contract.

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ExtendedMicroservicesContext.java
eutkin: The smaller the entities, the better. But I don't want to describe the contract in java doc, I think it's not enough. I have another idea to remove MicroservicesContext for ServiceFactory#getServiceDefinition. Because I don't see any cases for using MicroservicesContext

services-api/src/main/java/io/scalecube/services/ServiceFactory.java
artem-v: > What if we're reading from the disk?

We don't take in account theoretical arguments. Again. We don't need async mode to read a dozens of tuples such as Class+HasMap.

artem-v (changes requested):

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -309,17 +316,38 @@ public ServiceDiscovery discovery() {
...
   private Mono<Void> processBeforeDestroy() {
-    return Mono.whenDelayError(
-        methodRegistry.listServices().stream()
-            .map(ServiceInfo::serviceInstance)
-            .map(s -> Mono.fromRunnable(() -> Injector.processBeforeDestroy(this, s)))
-            .collect(Collectors.toList()));
+    return serviceFactory.shutdownServices(this.context).then();
+  }
+
+  public final class Context implements MicroservicesContext {
```
artem-v: Why it's public? Make it private.

artem-v (changes requested):

services-api/src/main/java/io/scalecube/services/ServiceProvider.java
artem-v: @eutkin @eutkin 
Coming back to this question.

Again, why those methods were deprecated:
```
services(Object... services)
services(ServiceInfo... services)
services(ServiceProvider serviceProvider)
```

Can't u keep them and use ScalecubeServiceFactory which will be defined internally? 
```
services(Object... services)
services(ServiceInfo... services)
services(ServiceProvider serviceProvider)
services(ServiceFactory serviceFactory)
```
I.e first three works with internal ScalecubeServiceFactory (client will not know even) + last one allows to specifiy explicitly serviceFactory. This way u would not need to deprecate you would just expose internal toolset called `ServiceFactory`.

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Used for remote service call.
+   *
+   * @return service call
+   * @see ServiceCall
+   */
+  ServiceCall serviceCall();
+
+  /**
+   * Network address of Scalecube node.
```
artem-v: Change to `Network address of Scalecube service discovery.`

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,43 @@
...
+ * @see ServiceFactory
+ */
+public interface MicroservicesContext {
+
+  /**
+   * Id of Scalecube node.
+   *
+   * @return id
+   */
+  String id();
```
artem-v: Expose `.serviceEndpoint()`  instead of `.id()`, service-endpoint along with `id` contains more useful information.

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Used for remote service call.
```
artem-v, line 24: Change to `Returns new instance {@code ServiceCall}.`

services-api/src/main/java/io/scalecube/services/MicroservicesContext.java
```diff
@@ -0,0 +1,43 @@
...
+  Address serviceAddress();
+
+  /**
+   * Service discovery for services localed in other nodes.
+   *
+   * @return service discovery
+   * @see ServiceDiscovery
+   */
+  ServiceDiscovery serviceDiscovery();
+}
```
artem-v, line 46: Let's not expose entire `ServiceDiscovery` (with `.start()` and `.stop()`) -- expose only what's needed: 

`serviceEndpoint()`
`discoveryAddress()` 
`Flux<ServiceDiscoveryEvent> listenDiscoveryEvents()`

services-api/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -367,7 +367,7 @@ public ServiceCall contentType(String contentType) {
     return (T)
         Proxy.newProxyInstance(
             getClass().getClassLoader(),
-            new Class[] {serviceInterface},
+            new Class[]{serviceInterface},
```
artem-v, line 370: Fix formatting

services-examples-parent/services-examples/src/main/java/io/scalecube/services/examples/exceptions/ExceptionMapperExample.java
```diff
@@ -31,24 +40,34 @@ public static void main(String[] args) throws InterruptedException {
...
+          public Mono<? extends Collection<ServiceInfo>> initializeServices(
+              MicroservicesContext microservices) {
+            ServiceCall call = microservices.serviceCall();
+            ServiceA serviceA =
+                call.errorMapper(new ServiceAClientErrorMapper()).api(ServiceA.class);
+            ServiceB serviceB = new ServiceBImpl(serviceA);
+            ServiceInfo serviceInfoB = ServiceInfo.fromServiceInstance(serviceB).build();
+            return Mono.just(Collections.singletonList(serviceInfoB));
+          }
+        };
```
artem-v, line 58: Just realized how `ServiceFactory` api can shoot in the foot. In this example with service-a service-b in method `getServiceDefinitions` were returned two services, in method `initializeServices` was returned one service. That's ok that service-b injects service-a, but i imagine scenario with N services and now client has to maintain correlation bewtween what's returned from `getServiceDefinitions` and `initializeServices`. 
Don't have solution for this, but to me it looks like a big problem.

artem-v (changes requested):

services/src/test/java/io/scalecube/services/sut/GreetingServiceImpl.java
artem-v: @ronenhamias @segabriel btw question about _ExtendedMicroservicesContext_ is not relevant anymore, it will be one `MicroservicesContext`.

services-examples-parent/services-examples/src/main/java/io/scalecube/services/examples/exceptions/ExceptionMapperExample.java
eutkin: There are only two ways. 

1. We create instances of services from ServiceDefinition
2. 

It is assumed that ServiceFactory will delegate its task to the DI framework and there will be no troubles. But they are possible. I will think about it.

eutkin:
@ronenhamias @segabriel @artem-v I'm extract all public API on `Microservices` to `MicroserviceContext`. Do we have any other major problems that are preventing me from merging my PR? Merge with develop branch takes a lot of time, as do endless edits.

eutkin:
up

eutkin:
What other important problems remain unresolved? It's becoming more and more difficult to merge this PR with the develop branch.

eutkin:
@ronenhamias @artem-v @segabriel 

When I was writing the documentation, I found a semantic mistake. I suggest changing `ServiceFactory` [in this way](https://gist.github.com/eutkin/e29e3d2e0c26665a1202bdbfa205b7a3).

Why? Because otherwise it seems as if we are managing the life cycle of only those services that will be placed on the Scalecube node. In fact, ServiceFactory is needed to manage the lifecycle of all user components. But unlike DI frameworks `ServiceFactory` separates Scalecube services from all components.

segabriel:
@eutkin 
> What other important problems remain unresolved? It's becoming more and more difficult to merge this PR with the develop branch.

I guess we need to postpone this PR for a week probably two weeks, due to we need to have a stable version of scalecube-services for this time slot. And have the ability to fast solve some shortcomings that stop us.

PR #716, opened 2020-04-25
Add functionality for custom codecs installation

segabriel:
Now to send something with some custom codecs we can't use proxy api and it requires to use `ServiceMessage` like:

```java
ServiceMessage request = ServiceMessage.builder()
            .qualifier(qualifier)
            .dataFormat("custom-content-type")
            .data(data)
            .build()
```

But it's not enough to send back a response in the same custom content type we should use `ServiceMessage` and specify incoming content-type again:

```java
ServiceMessage response = ServiceMessage.builder()
            .qualifier(qualifier)
            .dataFormat("custom-content-type")
            .data(data)
            .build()
```

 Sum up: proxy API can not work with any custom codec except for the hard-coded `application/json` and having this codec in the classpath is mandatory. And the current PR tries to remove these shortcomings.

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/ServiceCall.java
+++ b/services-api/src/main/java/io/scalecube/services/ServiceCall.java
@@ -48,6 +48,7 @@ public class ServiceCall {
         // no-op
       };
   private Map<String, String> credentials = Collections.emptyMap();
+  private String contentType;
 
   /** Default constructor. */
   public ServiceCall() {}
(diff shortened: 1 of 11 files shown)


artem-v (changes requested):
https://github.com/scalecube/scalecube-services/pull/717/files

artem-v (approved):
thanks @segabriel !

PR #730, opened 2020-05-10
Update development md

aharonha:
* using java 8 (openjdk)
* small typo


artem-v:
@aharonha 
we're on 11 everywhere 
maybe let's update doc to 11 and drop 8 from docs

aharonha:
@artem-v  When did that happen? Why didn't you bump the version to 3?

artem-v:
> @artem-v When did that happen? Why didn't you bump the version to 3?

Which version to 3?

artem-v:
@aharonha 

> > @artem-v When did that happen? Why didn't you bump the version to 3?
> 
> Which version to 3?

Finally understood which version to `3`  . Let me clarify.

> we're on 11 everywhere

We compile against 11, this is true, but nowhere in scalecube-services use compile/runtime features of 11 java.


> maybe let's update doc to 11 and drop 8 from docs

No, let's not. 


> @artem-v When did that happen? Why didn't you bump the version to 3?

Again. We just compile against 11, but not introducing java-11-ness  into the code.

artem-v (approved):

aharonha:
Can one use the latest version of scalecube services on java 8?
I think I have tried and I failed.
The binary uploaded to maven central repository has class version of 11 (55).
I might be wrong and I got a local compilation mixed up....

artem-v:
> Can one use the latest version of scalecube services on java 8?
> I think I have tried and I failed.
> The binary uploaded to maven central repository has class version of 11 (55).
> I might be wrong and I got a local compilation mixed up....

artem-v:
> Can one use the latest version of scalecube services on java 8?
> I think I have tried and I failed.
> The binary uploaded to maven central repository has class version of 11 (55).
> I might be wrong and I got a local compilation mixed up....

Create PR with changes in scalecube parent. I guess u need to tweak settings of maven compiler plugin (change target and source).

artem-v (approved):

artem-v:
@aharonha since 2.10.0 services are on jdk8.

PR #753, opened 2020-05-22
Add error type as a replacement for error qualifier and put original qualifier to error message

### LINKED ISSUE #751: Change error qualifer format
Migrate `io.scalecube.services.error/:errorType`  to  message with normal qualifier  but with header denoting error type. Leave ErrorData as is.

Add `_error_type`  header (name it by analog with `_data_format`).

Remove method `Qualfier.isError`, change inplementation of `ServiceMessage.isError`.

segabriel:
Closes #751

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/ServiceCall.java
+++ b/services-api/src/main/java/io/scalecube/services/ServiceCall.java
@@ -476,18 +476,13 @@ public class ServiceCall {
   }
 
   private Function<Flux<ServiceMessage>, Flux<Object>> asFlux(boolean isReturnTypeServiceMessage) {
-    return flux -> isReturnTypeServiceMessage ? flux.cast(Object.class) : flux.map(toResponse());
+    return flux ->
+        isReturnTypeServiceMessage ? flux.cast(Object.class) : flux.map(ServiceMessage::data);
   }
 
   private Function<Mono<ServiceMessage>, Mono<Object>> asMono(boolean isReturnTypeServiceMessage) {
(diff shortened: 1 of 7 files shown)


services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
```diff
@@ -169,12 +171,12 @@ public boolean isError() {
    * @return error type.
    */
   public int errorType() {
-    if (!isError()) {
+    final String errorType = headers.get(ERROR_TYPE);
+    if (errorType == null) {
       throw new IllegalStateException("Message is not an error");
     }
```
artem-v: Let's start returning `-1`. Let's change contract. I regret about this `throw new IllegalSt...`.

services-api/src/main/java/io/scalecube/services/ServiceCall.java
```diff
@@ -487,7 +484,10 @@ private ServiceUnavailableException noReachableMemberException(ServiceMessage re
   }
 
   private Function<ServiceMessage, Object> toResponse() {
-    return message -> message.hasData() ? message.data() : UNEXPECTED_EMPTY_RESPONSE;
+    return message ->
+        message.hasData()
+            ? message.data()
+            : ServiceMessage.error(message.qualifier(), 503, 503, "Unexpected empty response");
```
artem-v: Why this method exists et al? I understand that neither you (and me) created it, but do u remember by chance?

artem-v (changes requested):

artem-v (approved):
+1
thanks @segabriel

PR #759, opened 2020-06-05
Rename headers

### LINKED ISSUE #757: Headers Naming Convention
Currently header fields are look like `_type`, `_data_format` and `_error_type`.

`_error_type` is exposed to public and therefor need to align the headers to like all fields convention for example `_error_type` should be `errorType`.

segabriel:
Closes https://github.com/scalecube/scalecube-services/issues/757

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
+++ b/services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
@@ -28,11 +28,13 @@ public final class ServiceMessage {
   /** Error type header. */
   public static final String HEADER_ERROR_TYPE = "errorType";
 
-  private Map<String, String> headers = new HashMap<>(1);
+  private final Map<String, String> headers;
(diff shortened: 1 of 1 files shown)


artem-v (approved):

services-api/src/main/java/io/scalecube/services/api/ServiceMessage.java
```diff
@@ -171,7 +173,7 @@ public boolean isError() {
    * @return error type.
    */
   public int errorType() {
-    String errorType = headers.get(ERROR_TYPE);
+    String errorType = headers.get(HEADER_ERROR_TYPE);
     if (errorType == null) {
       return -1;
```
artem-v: Plz move `-1` to public constant called `NULL_ERROR_TYPE`, and udpated javadoc on `errorType()` method accordingly.

artem-v (approved):

PR #790, opened 2020-08-06
Fixed an error caused by Reflect.isRequestTypeServiceMessage() returning false on Publisher<ServiceMessage>

harry-hao:
When using dispatcher and REQUEST_CHANNEL together like follow:
```
  @ServiceMethod
  @RequestType(GreetingRequest.class)
  @ResponseType(GreetingResponse.class)
  Flux<ServiceMessage> bidiGreetingMessage(Publisher<ServiceMessage> requests);
```
When called it will have error `GreetingRequest cannot be cast to ServiceMessage`.
The error happens at `ServiceMethodInvoker::toRequest`:
```
return methodInfo.isRequestTypeServiceMessage() ? request : request.data();
```
In this case, it should return request, but the isRequestTypeServiceMessage() returns false, and request.data() is returned.

This PR tries to fix this by checking the actual type arguments, then returns true. 

Please take a look if this is expected behavior or should be fixed.

Thank you.

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/Reflect.java
+++ b/services-api/src/main/java/io/scalecube/services/Reflect.java
@@ -115,9 +115,18 @@ public class Reflect {
    * @return true if the first parameter of method is ServiceMessage, otherwise false
    */
   public static boolean isRequestTypeServiceMessage(Method method) {
-    Class<?>[] parameterTypes = method.getParameterTypes();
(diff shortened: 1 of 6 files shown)


harry-hao:
Closed to change the source branch, please see https://github.com/scalecube/scalecube-services/pull/791 instead.

PR #819, opened 2021-08-19
Update versions

segabriel:
- Update versions.
- Set the cluster transport and its metadata explicitly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,18 +57,18 @@
   </scm>
 
   <properties>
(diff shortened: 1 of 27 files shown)


artem-v (dismissed):
Plz  remove constructor `ScalecubeServicDiscovery(ServiceEndpoint)`, lets have default no-arg  constructor.  Lets have  `serviceEndpoint`  being set via unary-operaator  arppoach  via  .`options()`

PR #821, opened 2021-08-27
Allow subclasses to be considered as services.

matyasberry:
as the title and test says.

### DIFF
--- a/services-api/src/main/java/io/scalecube/services/Reflect.java
+++ b/services-api/src/main/java/io/scalecube/services/Reflect.java
@@ -18,11 +18,13 @@ import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


segabriel (approved):

ronenhamias:
can you provide example that showcase what is considered to be subclasses
something that shows useful usecase as reference for future users.
also explain the feature in the issue so it serve as a doc for discussion sake if someone will look it up in the future.

matyasberry:
> can you provide example that showcase what is considered to be subclasses

yes, it's in the enhanced test file:
```
@Service
  private interface SimpleService {
    public String name();
  }

  private class ServiceImpl implements SimpleService {
    @Override
    public String name() {
      return "duke";
    }
  }

  private class SubServiceImpl extends ServiceImpl {

  }

```

`SubServiceImpl` is a subclass which before this change was not considered (and detected) as a service implementation.

artem-v (approved):

PR #822, opened 2021-09-17
Always authenticate connections

josephtarbit:
If an authenticator is present, always authenticate connections regardless of whether credentials are present because at the moment if a client wants to bypass authentication they can just not specify any credentials.


artem-v:
It doesnt matter did you set  authenticator or have you  provided  credentials. Only  service method will judge.  Service method will  look for  AUTH_CONTEXT   and  if  it  is null  then caller will get error. 
One more. If  you  set  authenticator   and  provided  credentials  and  credentials  are valid   then  you're  allowed  to  est.   connection,  and  auth_context  will be set. But. Still  only  service method  will judge,  because  you  might  call method  which  is  not  allowed  to be called  for  your  credentials  with  which  you  successfully  est.  connection.
P.S
Ofcourse for this to happen service method must be defined with annotation `io.scalecube.services.auth.Secured`.

josephtarbit:
So every single service method needs to do it's own authentication checks?

artem-v:
> So every single service method needs to do it's own authentication checks?

If service method is annotated with `@Secured`  then - yes, otherwise - no.

PR #826, opened 2022-03-06
Build with Java 17 & Invoke default methods

josephtarbit:
This allows default methods on service interfaces to be called, although it requires at least Java 16 to function as it makes use of [InvocationHandler#invokeDefault](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/reflect/InvocationHandler.html#invokeDefault(java.lang.Object,java.lang.reflect.Method,java.lang.Object...)).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,8 @@
   </scm>
 
   <properties>
+    <maven.compiler.release>17</maven.compiler.release>
+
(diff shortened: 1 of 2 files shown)


josephtarbit:
Any updates on this?

PR #832, opened 2023-05-25
fix: authenticate ignored when the client is not set credentials

zhou-hao:
When the client is not set credentials,Authenticate is ignored,May lead to unauthorized requests?


artem-v:
Hello @zhou-hao, 
Any chance to get some test that reproduce this problem?

zhou-hao:
And when client set `ConnectionSetup` to `null`,The authentication logic will be invalid.

artem-v:
>The authentication logic will be invalid.

Can you be more specific - what exactly will be invalid?

zhou-hao:
Server side authenticator dose not work, when  client not set credentials.

https://github.com/zhou-hao/scalecube-services/blob/9618976cdcdbc820add9152b4bfeaf99af709801/services/src/test/java/io/scalecube/services/transport/rsocket/RSocketClientTransportTest.java#L32-L48
```java    

RSocketServiceTransport server = new RSocketServiceTransport()
      .serverTransportFactory(RSocketServerTransportFactory.tcp(serverPort))
      .clientTransportFactory(RSocketClientTransportFactory.tcp())
      // .credentialsSupplier((service) -> Mono.just(Collections.singletonMap(authKey, authValue)))
      // server side authenticator
      .authenticator(header -> {
        if (!Objects.equals(header.get(authKey), authValue)) {
          return Mono.error(new ForbiddenException("token invalid"));
        }
        return Mono.empty();
      });

RSocketServiceTransport client = new RSocketServiceTransport()
  .serverTransportFactory(RSocketServerTransportFactory.tcp(18811))
  .clientTransportFactory(RSocketClientTransportFactory.tcp())
  // client not set credentials
  //.credentialsSupplier((service) -> Mono.just(Collections.singletonMap(authKey, authValue)))
  ;
```

artem-v:
> Server side authenticator dose not work, when client not set credentials.

Why this is a problem? If service has annotation `@Secured` - then it doesn't matter did authenticator work or not, did client provide credentials or not - for secured service method only matters - an auth data in the context of a call - see `io.scalecube.services.methods.ServiceMethodInvoker#authenticate`

zhou-hao:
I thought this was a global enable authentication...  
Does it mean that every interface needs to be annotated `@Secured` ?

artem-v:
> I thought this was a global enable authentication... Does it mean that every interface needs to be annotated `@Secured` ?

There's  authenticator on Microsrvices.transport level - you set it when expect clients credentials on per-session level.  There's  authneticator on Microservices level + on ServiceInstance level  - those are service authneticators - they applied on per message level, you set them when expect  clients credentials on per-message level. 

> Does it mean that every interface needs to be annotated `@Secured` ?

Every service interface that assumes some protection either on per-session level or per-message level -  yes it must be annotated with `@Secured`.

zhou-hao:
> There's authenticator on Microsrvices.transport level - you set it when expect clients credentials on per-session level.

When authenticator on Microsrvices.transport level , Why is per-session  authentication not mandatory?

artem-v:
> > There's authenticator on Microsrvices.transport level - you set it when expect clients credentials on per-session level.
> 
> When authenticator on Microsrvices.transport level , Why is per-session authentication not mandatory?

In order to consolidate logic in one place - in `ServiceMethodInvoker#authenticate`.

PR #854, opened 2024-10-05
Better bootstrap process

artem-v:
Changes: 

- Less boilerplate code
- Less garbage and additional nonsense classes
- New (more short, more correct, less verbose) approach  of starting ` Microservices` object
- Get rid of `slf4j`.


aharonha (commented):
switching to 17 is great!

artem-v:
> switching to 17 is great!

WIP

services/src/main/java/io/scalecube/services/Microservices.java
```diff
@@ -114,283 +105,389 @@
  *
  * }</pre>
  */
-public final class Microservices implements AutoCloseable {
+public class Microservices implements AutoCloseable {
 
-  public static final Logger LOGGER = LoggerFactory.getLogger(Microservices.class);
+  // private static final Logger LOGGER = LoggerFactory.getLogger(Microservices.class);
```
segabriel: Polish the code first.

PR #983, opened 2026-09-22
JDK25 update

matyasberry:
JDK25 update


segabriel (approved):
Don't forget to update the versions in the properties block.

